PR #4, opened 2018-02-21
NDCUBE to the rescue 

nabobalis:
We can use this as a test bed for pycodestyle with the aim of getting it online for SunPy 1.0


nabobalis:
Ima rebase this mess.

PR #8, opened 2018-02-23
Some initial functionality

wafels:
This PR contains some initial functionality, and is based on an expanded version of the [offlimb enhancement example in the sunpy gallery](https://github.com/sunpy/sunpy/blob/master/examples/off_limb_enhance.py).  This PR also contains an implementation of the normalizing radial gradient filter of [Morgan, Habbal & Woo, 2006, Sol. Phys., 236, 263](https://link.springer.com/article/10.1007%2Fs11207-006-0113-6).


prateekiiest:
@wafels, could I get to see the IDL output of this implementation?

wafels:
The function intensity_enhance does not have an equivalent IDL implementation in SolarSoft, as far as I know.  

There is an implementation of the NRGF, and it is here:

http://heliodocs.com/php/xdoc_print.php?file=$SSW/packages/nrl/idl/nrlgen/display/nrgf.pro

However, this implementation includes a *lot* more extra stuff which is not described in the paper like looking for bad pixels before doing the processing.  Functionality like this should be separated from the implementation of Equation (1) in the paper https://arxiv.org/pdf/astro-ph/0602174.pdf .  The function "normalizing_radial_gradient_filter" in the PR is intended to implement Equation (1) of the paper, and nothing else.

So I'd say that there is no Solarsoft/IDL equivalent of the implementations in the PR.

wafels:
Closing PR so as not to waste CI cycles.  Please continue to work on the code as need be!

PR #12, opened 2018-07-25
Noise level estimation for a single image. 

MSKirk:
A direct translation of NoiseLevel.m to estimate AWGN using weakly textured patches.


DanRyanIrish:
```NoiseLevel_test.py``` should be ```test_noiselevel.py``` or ```test_noise_level.py```

File: sunkit_image/utils/NoiseLevel.py [hunk 1]
```diff
@@ -0,0 +1,201 @@
+import numpy as np
+from scipy.stats import gamma
+from scipy.ndimage import correlate
+from skimage.util import view_as_windows
+
```
Cadair: we should have an `__all__` list with the function / class names here.

sunkit_image/utils/NoiseLevel.py
```diff
@@ -0,0 +1,201 @@
...
+
+class NoiseLevelEstimation:
+    def __init__(self, img, patchsize=7, decim=0, conf=1-1E-6, itr=3):
+        """
+        NoiseLevel estimates noise level of input single noisy image.
+
+        Input parameters:
```
Cadair: This should use numpydoc styling, i.e.
```
Parameters
----------
```

same as the other section headings :)

sunkit_image/utils/NoiseLevel.py
```diff
@@ -0,0 +1,201 @@
...
+
+        Example:
+            estimate = NoiseLevelEstimation(noisy_image_array, patchsize=11, itr=10)
+
+        Python Version: 20180718
+        Python Author: M. Kirk
+
+        Translated from Noise Level Estimation Matlab code: noiselevel.m
+
+        noiselevel.m Copyright (C) 2012-2015 Masayuki Tanaka
```
Cadair: What licence is the original version under?

sunkit_image/utils/NoiseLevel.py
```diff
@@ -0,0 +1,201 @@
...
+            itr (optional): number of iteration. (default: 3)
+
+        Calculated parameters:
+            nlevel: estimated noise levels.
+            th: threshold to extract weak texture patches at the last iteration.
+            num: number of extracted weak texture patches at the last iteration.
+            mask: weak-texture mask. 0 and 1 represent non-weak-texture and weak-texture regions, respectively
+
+        Example:
+            estimate = NoiseLevelEstimation(noisy_image_array, patchsize=11, itr=10)
```
Cadair: Add an example of accessing the calculate parameters?

sunkit_image/utils/NoiseLevel.py
MSKirk: From how I read it, I think we can include it as long as we add in the original license file.

Cadair: looks like BSD 2-clause to me, I agree that we can use it.

wafels:
I think this should live in its own subdirectory, something like /noise.  Is there a descriptive name for the method we could use instead of "NoiseLevel".  Also, what does AWGN mean?

MSKirk:
"Additive white Gaussian noise (AWGN) is a basic noise model used in Information theory to mimic the effect of many random processes that occur in nature."

nabobalis:
Would it be ok If I rebased this and addressed the comments?

MSKirk:
@nabobalis Go for it! I haven't had the time to look at it recently, but I intend to get back to that project soon.

nabobalis:
My documentation will need a read over to make sure it makes any sense.

nabobalis:
Thanks @MSKirk!

MSKirk:
Thanks @nabobalis for pushing this across the finish line.

nabobalis:
You did the hard work, I just tidied some strings up!

PR #15, opened 2019-01-15
Making sure the CI passes.

nabobalis:
See title.

Todo:

- [ ] Fix this.

### DIFF
--- a/sunkit_image/utils/tests/test_utils.py
+++ b/sunkit_image/utils/tests/test_utils.py
@@ -6,6 +6,7 @@ import astropy.units as u
 from astropy.tests.helper import assert_quantity_allclose
 
 import sunpy.map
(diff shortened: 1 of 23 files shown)


sunkit_image/utils/tests/test_utils.py
```diff
@@ -17,6 +17,7 @@
 
 @pytest.fixture
 def smap():
+    import sunpy.data.sample
```
Cadair: the horror

nabobalis: Yeah I want to check something but the tox env wont install on my local machine.

nabobalis: Doesn't work. I can't see to get the sample data downloaded for the test. I have to come up with another method.

Issue #18, opened 2019-03-03
De-Class NoiseLevelEstimation


nabobalis:
### Description

When #12 is merged in, I propose turning `NoiseLevelEstimation` into a series of functions and not a class.

MSKirk:
I agree on this front. There are several noise level estimation routines that are appropriate for use in solar images. I think a series of functions would better represent these.

m-squared96:
Hey, just a heads up. I'm a GSoC hopeful and I'm attempting to fix this issue. Currently WIP. I have the original NoiseLevelEstimation object "de-classed" and now I just need to test to make sure functionality has remained the same.

dpshelio:
I would take a different approach. I would have a denoise class and different functions as different strategies, that is how the strategy pattern works. This is the implementation on a [solar physics example](http://github-pages.ucl.ac.uk/rsd-engineeringcourse/ch05construction/09patterns.html#Strategy-Pattern).

nabobalis:
So we just refactor the class to enable someone to add more functions?

PR #21, opened 2019-03-23
fixes issue #18 - Converted class into series of functions

ujjwalll:
### Description

This pull request coverts the noise.py file from class to series of functions and appropriately adjusts the test_noise.py file

Fixes #<Issue Number>

### DIFF
--- a/sunkit_image/utils/noise.py
+++ b/sunkit_image/utils/noise.py
@@ -8,12 +8,27 @@ from skimage.util import view_as_windows
 from scipy.ndimage import correlate
 
 __all__ = ["NoiseLevelEstimation"]
-
 img = 0
-patchsize=7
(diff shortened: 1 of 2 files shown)


nabobalis:
Thanks @ujjwalll, however I think you have slightly misunderstood what was meant by removing the class. The idea was that each new function would be standalone, as currently stands you have not really removed the class, just reformed it.

ujjwalll:
@nabobalis At present, the functions in noise.py are dependent on initialize function, I will remove that so that every function will become a standalone function. I will initialize variables in every function according to need.

nabobalis:
I think this can be closed by #22 as I think that is a better place to be merged. Sorry about this @ujjwalll.

PR #24, opened 2019-04-08
Added NAFE

### LINKED ISSUE #6: Noise Adaptive Fuzzy Equalization method (NAFE)
The idea is from this [paper.](http://iopscience.iop.org/article/10.1088/0067-0049/207/2/25)

The original attempt is located here (https://github.com/sunpy/sunpy/pull/1899).
This needs to be moved into this module and have the following done to it:

- [ ] Test to compare to known output ([original version](http://www.zam.fme.vutbr.cz/~druck/Nafe/Index.htm)).
- [ ] Optimized as much as possible for memory and CPU time.
- [ ] 100% test coverage
- [ ] Documentation and a worked example.
- [ ] Compared to another version in [Python.](https://git.ias.u-psud.fr/ebuchlin/aia-movie/blob/master/medocimage/nafe.py) Maybe both can exist within this package.

MinaSGorgi:
### Description

This is a complete working implementation of the [NAFE](https://iopscience.iop.org/article/10.1088/0067-0049/207/2/25) filter. Only missing feature is setting default values for input and output margins according to image contrast which is provided by [original version](http://www.zam.fme.vutbr.cz/~druck/Nafe/Index.htm). Otherwise, it supports all features including default parameters for different wavelengths. Original code is not open-sourced, hence can't compare it code-wise and thorough testing against original code results is still in progress. Differences between this implementation and other available [open-source version](https://git.ias.u-psud.fr/ebuchlin/aia-movie/blob/master/medocimage/nafe.py) are mainly implementation-wise. The only non-trivial difference is using constant noise removal sigma.

An output sample: (400x400) in ~10 mins
Screenshot from 2019-04-08 05-56-46

Creating this PR for feedback. @nabobalis @dpshelio 

Fixes #6

### DIFF
--- a/sunkit_image/nafe.py
+++ b/sunkit_image/nafe.py
@@ -1,0 +1,282 @@
+"""
+This module contains functions that can be used to enchance tof fine structures in extreme
+ultraviolet images of the corona.
+"""
+import argparse
+import warnings
+import multiprocessing
+from collections import namedtuple
+from multiprocessing import Pool
+
+import tqdm
+import numpy as np
+import skimage.filters
+
+import sunpy.map
+
+__all__ = [
+    "nafe"
+]
+
+
+_PREDEF_PARAMS = {
+    # defined by original program
+    # wavelength    # gamma     # nafe
+    "4500":          (1.0,        .05),
+    "1700":          (2.2,        .10),
+    "1600":          (2.2,        .10),
+    "335":           (3.0,        .20),
+    "304":           (5.5,        .15),
+    "211":           (2.7,        .20),
+    "193":           (2.6,        .20),
+    "171":           (2.6,        .20),
+    "131":           (2.9,        .20),
+    "94":            (3.1,        .20),
+    "def":           (2.2,        .20)
+}
+
+_SharedParams = namedtuple("_SharedParams", ["image", "membership", "hist_bins", "gamma", "sigma",
+                                             "weight", "in_margin", "halfn", "out_margin"])
+_SHARED_PARAMS = None  # for sharing the image and parameters across processes
+
+
+def nafe(input_map, in_margin=None, out_margin=(0, 2 ** 16 - 1), gamma=None, nafe_weight=None,
+         hist_bins= 2 ** 8 - 1, noise_reduction_sigma=15, n=129, c=np.ones(12), nproc=None):
+    """
+    Enhances the input map via "Noise Adaptive Fuzzy Equalization". Ideally suited for the
+    visualization of fine structures in extreme ultraviolet images of the corona. Particularly
+    suited for the exceptionally high dynamic range images from the Atmospheric Imaging Assembly
+    instrument on the Solar Dynamics Observatory. This method produces artifact-free images and
+    gives significantly better results than methods based on convolution or Fourier transform which
+    are often used for that purpose.
+
+    Parameters
+    ----------
+    input_map: `~sunpy.map.Map`
+        The original input map to be enhanced.
+    in_margin: `tuple`
+        (low, high) input image values. Pixels with values < low or > high are set to  0.
+    out_margin: `tuple`
+        (low, high) output image values. Deaults to (0, 2 ** 16 - 1).
+    gamma: `float`
+        The gamma value used for gamma_transform.
+    nafe_weight: `float`
+        Controls the ratio of the nafe component in the result image. Typical values for nafe_weight
+        lie in the interval [0.05, 0.3].
+    hist_bins: `int`
+        Number of histogram bins used for calculating fuzzy historgam for nafe component. Defaults
+        to 2 ** 8 - 1.
+    noise_reduction_sigma: `float`
+        Sigma of artificial gaussian noise added to the input image for nafe component. Defaults to
+        15.
+    n: `int`
+        Full kernel width used for calculating neighborhood for nafe component. Must be an odd
+        integer. Defaults to 129.
+    c: `~numpy.ndarray`
+        Fuzzy neighborhood constants for nafe component. Defaults to 12 layers each of coefficient
+        1.
+    nproc : `int`
+        Number of processes for multiprocessing. Defaults to number of available cpu cores.
+
+    Returns
+    -------
+    enhanced_map: `~sunpy.map.Map`
+        The enhanced map resulting from the linear combination of two images,
+        gamma_transform(input_map) and nafe_transform(input_map) such that:
+
+            enhanced_map = (1 - nafe_weight) * gamma_component + nafe_weight * nafe_component
+
+    References
+    ----------
+
+    * `A Noise Adaptive Fuzzy Equalization Method for Processing Solar Extreme Ultraviolet Images <https://doi.org/10.1023/A:1005226402796>`__
+
+    """
+    global _SHARED_PARAMS
+    data = input_map.data
+    metadata = input_map.meta
+    predef_gamma, predef_weight = _PREDEF_PARAMS.get(metadata["wavelnth"], _PREDEF_PARAMS["def"])
+
+    if in_margin is None:
+        in_margin = (data.min(), data.max())
+    else:
+        data = data.clip(min=in_margin[0], max=in_margin[1])
+
+    if gamma is None:
+        gamma = predef_gamma
+    elif gamma < 0:
+        raise ValueError("gamma should be a non-negative real number! gamma given is:", gamma)
+
+    if nafe_weight == 0:
+        warnings.warn("nafe_weight is set to zero which disables nafe", RuntimeWarning)
+        enhanced_image = _transform(in_margin=in_margin, out_margin=out_margin, old_value=data,
+                                    power=gamma)
+    else:
+        if nafe_weight is None:
+            nafe_weight = predef_weight
+        elif nafe_weight < 0 or nafe_weight > 1:
+            raise ValueError("nafe_weight should be between 0 and 1. nafe_weight given is:",
+                             nafe_weight)
+        elif nafe_weight < 0.05 or nafe_weight > 0.3:
+            warnings.warn("typical values for nafe_weight should be between 0.05 and 0.3! "
+                          "nafe_weight given is: " + str(nafe_weight), RuntimeWarning)
+
+        if not isinstance(hist_bins, int) or hist_bins <= 0:
+            raise ValueError("hist_bins should be a positive integer number! hist_bins given is:",
+                             hist_bins)
+
+        if noise_reduction_sigma < 0.0:
+            raise ValueError("noise_reduction_sigma should be a non-negative real number! "
+                             "noise_reduction_sigma given is:", noise_reduction_sigma)
+
+        if n <= 0 or n % 2 == 0 or not isinstance(n, int):
+            raise ValueError("n should be an odd positive integer number! n given is:", n)
+
+        if nproc is None:
+            nproc = multiprocessing.cpu_count()
+        elif not isinstance(nproc, int) or nproc <= 0:
+            raise ValueError("nproc should be a non-negative integer number! nproc given is:",
+                             nproc)
+        elif nproc > multiprocessing.cpu_count():
+            raise ValueError("nproc given is too big! Maximum value should be:",
+                             multiprocessing.cpu_count())
+
+        # Shared nafe parameters
+        membership = _get_membership_grade(n=n, wl=c)
+        halfn = n // 2
+        _SHARED_PARAMS = _SharedParams(image=data, membership=membership, hist_bins=hist_bins,
+                                       gamma=gamma, sigma=noise_reduction_sigma, weight=nafe_weight,
+                                       in_margin=in_margin, halfn=halfn,
+                                       out_margin=(0, 2 ** 16 - 1))
+
+        # NAFE multiprocessing context
+        nrows, ncols = data.shape
+        size = nrows * ncols
+        rows, cols = np.meshgrid(range(nrows), range(ncols), indexing='ij')
+        indices = zip(rows.flat, cols.flat)
+        with Pool(processes=nproc) as pool:
+            # return margin [0, 2 ** 16 - 1]
+            nafe_image = list(tqdm.tqdm(pool.imap(_nafe_slice, indices), total=size))
+
+        enhanced_image = np.array(nafe_image).reshape(nrows, ncols)
+        enhanced_image = _transform(in_margin=(0., 2. ** 16 - 1.), out_margin=out_margin,
+                                    old_value=enhanced_image)
+
+    return sunpy.map.Map((enhanced_image, metadata))
+
+
+def _get_membership_grade(n=129, wl=np.ones(12)):
+    """
+    Calculates the matrix used for expressing the membership grade of neighborhood pixels to the
+    fuzzy histogram computed for each pixel in image. The matrix L is a linear combination of
+    M Gaussian kernels calculated as:
+
+        L(x, y) = d * sum[m=1:M](wl[m] * G(sigma[m], x, y))
+
+    where:
+    x = horizontal distance from center
+    y = vertical distance from center
+    sigma[m] = 2 ** (m / 2)
+    M = len(wl)
+
+    d = 2 * pi / sum[m=1:M]( wl[m] * (sigma[m] ** - (1/2)) )
+    G(s, x, y) = exp(−0.5 * (x ** 2) + (y ** 2)) * (s ** -2)) / (2 * pi * sqrt(s)), s = sigma[m]
+
+    Parameters
+    ----------
+    n: `int`
+        Full neighborhood width. Must be an odd integer and defaults to 129.
+    wl: `~numpy.ndarray`
+        Weights list of for each Gaussian kernel of the M Gaussian kernels
+        used for computing L. M is calculated as len(wl). Defaults to a
+        1x12 matrix of ones.
+
+    Returns
+    -------
+    L: `~numpy.ndarray`
+        Membership grade matrix.
+
+    Example
+    -------
+    >>> _get_membership_grade(5)
+    array([[0.65289406, 0.73715583, 0.7741782 , 0.73715583, 0.65289406],
+           [0.73715583, 0.86813217, 0.92826105, 0.86813217, 0.73715583],
+           [0.7741782 , 0.92826105, 1.        , 0.92826105, 0.7741782 ],
+           [0.73715583, 0.86813217, 0.92826105, 0.86813217, 0.73715583],
+           [0.65289406, 0.73715583, 0.7741782 , 0.73715583, 0.65289406]])
+    """
+    M = len(wl)
+    sigma = [2 ** (m / 2) for m in range(1, M+1)]  # starts from 1 for maths
+    d = 2 * np.pi / np.sum(wl * np.power(sigma, -1/2))
+
+    wlG = 0
+    for s, w in zip(sigma, wl):
+        G = lambda r, c: np.exp(
+                -0.5 * ((r - n//2) ** 2 + (c - n//2) ** 2) * (s ** -2)) / (2 * np.pi * np.sqrt(s))
+        wlG += w * np.fromfunction(G, (n, n))
+
+    return d * wlG
+
+
+def _transform(in_margin, out_margin, old_value, power=1):
+    """
+    Gamma + Linear transform
+    """
+    in_low = in_margin[0]
+    out_low = out_margin[0]
+    in_range = in_margin[1] - in_margin[0]
+    out_range = out_margin[1] - out_margin[0]
+    new_value = (old_value - in_low) / in_range
+
+    return out_low + out_range * (new_value ** (1 / power))
+
+
+def _nafe_slice(index):
+    """
+    Core NAFE function running on one pixel for multiprocessing
+    """
+    global _SHARED_PARAMS
+    image = _SHARED_PARAMS.image
+    membership = _SHARED_PARAMS.membership
+    hist_bins = _SHARED_PARAMS.hist_bins
+    gamma = _SHARED_PARAMS.gamma
+    sigma = _SHARED_PARAMS.sigma
+    weight = _SHARED_PARAMS.weight
+    in_margin = _SHARED_PARAMS.in_margin
+    halfn = _SHARED_PARAMS.halfn
+    out_margin = _SHARED_PARAMS.out_margin
+
+    row, col = index
+    rows, cols = image.shape
+
+    img_minrow = max(0, row - halfn)
+    img_mincol = max(0, col - halfn)
+    img_maxrow = min(rows, row + halfn + 1)
+    img_maxcol = min(cols, col + halfn + 1)
+
+    mem_minrow = img_minrow - row + halfn
+    mem_mincol = img_mincol - col + halfn
+    mem_maxrow = img_maxrow - row + halfn
+    mem_maxcol = img_maxcol - col + halfn
+
+    neighborhood = image[img_minrow: img_maxrow, img_mincol: img_maxcol]
+    membership = membership[mem_minrow: mem_maxrow, mem_mincol: mem_maxcol]
+
+    hist, bins = np.histogram(neighborhood, bins=hist_bins, weights=membership)
+    sigma = sigma / ((bins[-1] - bins[0]) / hist_bins)
+    bins = (bins[:-1] + bins[1:]) / 2
+
+    cumsum_hist = np.cumsum(hist)
+    cumsum_hist /= cumsum_hist[-1]  # normalize
+    # add artificial noise, reshaping to dummy 2d to enable 2d gauss from skimage then
+    # flattening again
+    cum_fuzzy_hist = skimage.filters.gaussian(cumsum_hist.reshape(hist_bins, 1), sigma=sigma)
+    cum_fuzzy_hist = cum_fuzzy_hist.flat
+
+    old_val = image[row, col]
+    gamma_val = _transform(in_margin=in_margin, out_margin=out_margin, old_value=old_val,
+                           power=gamma)
+    nafe_val = np.interp(image[row, col], bins, cum_fuzzy_hist)  # returns [0., 1.]
+    nafe_val = _transform(in_margin=(0., 1.), out_margin=out_margin, old_value=nafe_val)
+
+    return (1 - weight) * gamma_val + weight * nafe_val
--- a/changelog/24.feature.rst
+++ b/changelog/24.feature.rst
@@ -1,0 +1,2 @@
+Added a module (`sunkit_image.nafe`) for enhancing fine structures in extreme ultraviolet images of
+the corona via "Noise Adaptive Fuzzy Equalization".
--- a/docs/code_ref/index.rst
+++ b/docs/code_ref/index.rst
@@ -8,5 +8,6 @@ Code Reference
    :maxdepth: 2
 
    sunkit_image
+   nafe
    offlimb_enhance
    utils
--- a/docs/code_ref/nafe.rst
+++ b/docs/code_ref/nafe.rst
@@ -1,0 +1,8 @@
+****
+nafe
+****
+
+This package implements Noise Adaptive Fuzzy Equalization for the visualization of fine structures
+in extreme ultraviolet images of the corona.
+
+.. automodapi:: sunkit_image.nafe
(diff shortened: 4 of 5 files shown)


nabobalis:
I will have a proper review later on but any chance you can change your image to focus on the edges of the Sun? I'd like to see the output on a patch of structures.

MinaSGorgy:
I get what you want but current sample images I am working with doesn't have such feature. If you can point a specific image in mind I can definitely provide that. This is the best I have currently.
Screenshot from 2019-04-08 08-24-12

NB: I currently remove all pixels out of ROI while instead I should keep them but evaluate new pixels for only the ones inside ROI. Will change locally for now.

nabobalis:
That region works, are you doing any extra normalization to plot the image?

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+  abstract={A new image enhancement tool ideally suited for the visualization
+  of fine structures in extreme ultraviolet images of the corona is presented
+  in this paper. The Noise Adaptive Fuzzy Equalization method is particularly
+  suited for the exceptionally high dynamic range images from the Atmospheric
+  Imaging Assembly instrument on the Solar Dynamics Observatory. This method
+  produces artifact-free images and gives significantly better results than
+  methods based on convolution or Fourier transform which are often used for
+  that purpose.}
+}
+"""
```
nabobalis: We don't need this starting string, if you want to add a reference, I would suggest adding to the main docstring of the function, as a bullet point. You should see an example in the noise.py file.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+def _transform(old_margin, new_margin, old_value, power=1.0):
+    """
+    # TODO: add doc
+    """
+    return new_margin[0] + (new_margin[1] - new_margin[0]) * (((old_value - old_margin[0]) / (old_margin[1] - old_margin[0])) ** (1.0 / power))
+
+
+class _Nafe():
```
nabobalis: I am not sure I am in favour of having a  class here, I would rather it be a series of functions with the "main" function that will what the user calls.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+                      gb=args["global_black"], lb=args["local_black"],
+                      gw=args["global_white"], lw=args["local_white"],
+                      n=args["neighborhood_width"], c=args["constants"])
+
+    figure, axes = plt.subplots(1, 2)
+    axes[0].imshow(input_map.data[1000:1400, 1000:1400], cmap='gray')
+    axes[0].set_title('Original Image')
+    axes[1].imshow(nafe_image, cmap='gray')
+    axes[1].set_title('Enhanced Image')
+    plt.show()
```
nabobalis: This can all go. We don't want to be able to call the function like this. All those `arser.add_argument(` should the parameters in the corresponding doc strings. 

This could be turned into a seperate gallery example.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    "131":           (2.9,        .20),
+    "94":            (3.1,        .20),
+    "def":           (2.2,        .20)
+}
+
+
+def nafe(input_map, imargin=None, gamma=None, omargin=None, nafe=False,
+         nafe_weight=None, roi_p=None, roi_f=None, nproc=None, hbins=None,
+         nrs=15, gb=0.3, lb=0.7, gw=0.3, lw=0.7, n=129,
+         c=[1 for _ in range(12)]):
```
nabobalis: This just creates a list of 1 that is 12 long?

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    high dynamic range images from the Atmospheric Imaging Assembly instrument
+    on the Solar Dynamics Observatory. This method produces artifact-free
+    images and gives significantly better results than methods based on
+    convolution or Fourier transform which are often used for that purpose.
+
+    Parameters
+    ----------
+    input_map : `~sunpy.map.sources.sdo.AIAMap`
+        The original input image to be enhanced.
+
```
nabobalis: You should remove these line breaks between parameters.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    imargin: `~tuple`
```
nabobalis: The `~` shorterns the name in the sphinx build, you only want it for long import. For the basic python types, you don't need them.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    Enhances image "input_map" via "Noise Adaptive Fuzzy Equalization"
+    method for better visualization. Particularly suited for the exceptionally
+    high dynamic range images from the Atmospheric Imaging Assembly instrument
+    on the Solar Dynamics Observatory. This method produces artifact-free
+    images and gives significantly better results than methods based on
+    convolution or Fourier transform which are often used for that purpose.
+
+    Parameters
+    ----------
+    input_map : `~sunpy.map.sources.sdo.AIAMap`
```
nabobalis: This will only work on `AIAMap` and not any `Map`?

sunkit_image/nafe.py
MinaSGorgy: It's basically for debugging purposes but will apply for final PR.

MinaSGorgy:
Thanks @nabobalis for your review and feedback. Regarding extra normalization, nothing not mentioned in the algorithm.

nabobalis:
In one of the examples in the SunPy Gallery, we apply just a log scale to the intensity. Any chance for the second example, you apply that to both images?

MinaSGorgy:
Applying logNorm on the last 2 images gives:
Screenshot from 2019-04-08 20-50-31

sunkit_image/nafe.py
MinaSGorgy: I am curious why would a class not be preferred. I come from a C++ background so that's why it seemed the way to go for me.

nabobalis: I guess I don't understand why we need a class here.

MinaSGorgy: Basically used for keeping reference of the shared variables across the processes pool. Just making the code more readable and the mapping function cleaner.

nabobalis: Is there no other way?

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+        if gw < 0 or lw < 0 or gw + lw != 1:
+            raise ValueError("Invalid values for (gw, lw)!", (gw, lw))
+
+        if not isinstance(n, int) or n % 2 == 0:
+            raise ValueError("n should be an odd integer. n given is:", n)
+
+        if len(c) <= 0:
+            raise ValueError("No constants(c) given! "
+                             "At least one should be given")
+
```
nabobalis: This is a large amount of input checking, is this all necessary?

MinaSGorgy: Not really sure how it's done, some basic checks are required at least like n being positive integer and gamma being non-negative real.

sunkit_image/nafe.py
MinaSGorgy: Pretty sure yes but that one liner mapping would be a lot messier.

MinaSGorgy: This: `pool.imap(nafe_instance.transform, indices), total=map_data.size)` is what I mean. I need to use imap because the order of the results matter since they are pixels values. AFAIK the only other way to pass the arguments to imap would be creating an iterable whose elements will mainly be in the form (row, col, some constant arguments). I thought since they are constant arguments this seemed a typical case for using a class.

sunkit_image/nafe.py
MinaSGorgy: Should I change this to take a numpy array instead and an optional wavelength argument? The original program uses the wavelength to calculate the optimal values for some other parameters. I can assign the default values for these parameters instead if not given a wavelength.

sunkit_image/nafe.py
MinaSGorgy: One of the reasons I went for this project was learning to develop high quality python code. My reference code-bases tended to do some input checking but I may have overdid it that's why I am not really sure.

sunkit_image/nafe.py
nabobalis: That would be a good idea.

sunkit_image/nafe.py
nabobalis: Well if it's the best way to do it, it can stay.

sunkit_image/nafe.py
nabobalis: I'm not against the idea but it seems very overkill. We should at least error on values that make zero  sense. For others, we can warn.

nabobalis:
Can you do a comparison with the other Python version?

nabobalis:
> Applying logNorm on the last 2 images gives:
> Screenshot from 2019-04-08 20-50-31

Thanks for that! Maybe it's just my old age, but I can't really spot any difference between these two images with the scaling now the same.

sunkit_image/nafe.py
dpshelio: I'm ok with the class, but why not having it all as one? I mean to include the `nafe` function as to execute it?

sunkit_image/nafe.py
dpshelio: Though I agree with @nabobalis and probably this wouldn't be the place. I imagine sunkit-image could have some command line programs that run some of these functions. This would need, however some extra additions to the `setup.py`.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+        nafe_val = _transform(old_margin=(0., 1.), new_margin=self.omargin, old_value=nafe_val)
+
+        return (1 - self.weight) * gamma_val + self.weight * nafe_val
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+
+    # Required
+    parser.add_argument("-i", "--image", help="path to input image", required=True)
```
dpshelio: normally, required arguments as the image in this case don't use a flag, but as a positional argument.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+from multiprocessing import Pool
+
+import sunpy.map
+
+import skimage
+from skimage import filters
+
+import matplotlib.pyplot as plt
+import numpy as np
+import tqdm
```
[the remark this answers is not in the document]
nabobalis: We can let isort work it out.

nabobalis:
Thanks for the review @dpshelio

sunkit_image/nafe.py
MinaSGorgy: Changed to previously mentioned example and will add link to the paper as in given example.

sunkit_image/nafe.py
MinaSGorgy: I have no preference but current implementation doesn't require a class anymore.

sunkit_image/nafe.py
MinaSGorgy: Will remove from file for now and investigate where to migrate after finishing testing.

sunkit_image/nafe.py
MinaSGorgy: Will apply the ones one.

sunkit_image/nafe.py
MinaSGorgy: I thought it would be easier to pass the map.data rather than possibly having to create a map from a numpy array. Again I have no preference. If a map would be sent I will check if its header contains info about the wavelength else I will use default value.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    "171":           (2.6,        .20),
+    "131":           (2.9,        .20),
+    "94":            (3.1,        .20),
+    "def":           (2.2,        .20)
+}
+
+
+def nafe(input_map, imargin=None, gamma=None, omargin=None, nafe=False,
+         nafe_weight=None, roi_p=None, roi_f=None, nproc=None, hbins=None,
+         nrs=15, gb=0.3, lb=0.7, gw=0.3, lw=0.7, n=129,
```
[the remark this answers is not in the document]
MinaSGorgy: Ideally the user would never change these parameters values. Some of them even aren't mentioned in the paper but I adopted them from the original software implemented by the paper's author. In this software they are even read from a file containing fine tuned default values. The author exposes changing them through changing the file, so I thought I would rather expose them as functions arguments. I totally agree with having self-explanatory variables as I normally go for but I took these as named in the mentioned file. Of course I can change if you find it better to.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+         c=[1 for _ in range(12)]):
+    """
+    Enhances image "input_map" via "Noise Adaptive Fuzzy Equalization"
```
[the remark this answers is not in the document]
MinaSGorgy: Will apply.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+            hbins = int(np.round((map_data.max() - map_data.min()) / h))
+        elif not isinstance(hbins, int) or hbins <= 0:
+            raise ValueError("hbins should be a non-negative integer number! "
+                             "hbins given is:", hbins)
+
+        if nrs < 0.0:
+            raise ValueError("nrs should be a non-negative real number! "
+                             "nrs given is:", nrs)
+
+        if gb < 0 or lb < 0 or gb + lb != 1:
```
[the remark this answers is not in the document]
MinaSGorgy: Oops my bad :sweat_smile: I guess they are misleading enough to change then.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+        rows, cols = np.meshgrid(range(nrows), range(ncols), indexing='ij')
+        indices = zip(rows.flat, cols.flat)
+
+        # NAFE multiprocessing context
+        with Pool(nproc) as pool:
+            # return margin [0, 2 ** 16 - 1]
+            nafe_image = list(tqdm.tqdm(pool.imap(nafe_instance.transform, indices), total=map_data.size))
+
+        enhanced_image = np.array(nafe_image).reshape(nrows, ncols)
+        enhanced_image = _transform(old_margin=(0., 2. ** 16 - 1.), new_margin=omargin, old_value=enhanced_image)
```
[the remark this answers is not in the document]
MinaSGorgy: Basically the value used by paper/original software. Need to check more into this tho.

sunkit_image/nafe.py
MinaSGorgy: Yes, missed this.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    convolution or Fourier transform which are often used for that purpose.
+
+    Parameters
+    ----------
+    input_map : `~sunpy.map.sources.sdo.AIAMap`
+        The original input image to be enhanced.
+
+    imargin: `~tuple`
+        (black, white) input image values. Pixels with values < black or
+        > white are set to  0.
```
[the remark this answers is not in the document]
MinaSGorgy: input. Yes and yes. Should be changed(as mentioned in a TODO) to a better contrast system as used by original software. The thing is, it's not mentioned anywhere.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+
+    gamma : `~float`
+        The gamma value used for gamma_transform.
+
+    omargin: `~tuple`
+        (black, white) output image values.
```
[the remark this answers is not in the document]
MinaSGorgy: Just using same terminology as original software. Can change of course.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+
+    nafe : `~bool`
+        Flag for enabling nafe transform and defaults to false.
```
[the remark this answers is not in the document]
MinaSGorgy: Adopted from original software since NAFE algorithm consists of 2 parts one is gamma correction and the other is the actual nafe (fuzzy histogram) which is computationally intensive hence not enabled (the nafe part) by default. I agree this is actually counter-intuitive and will change to default to true.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+
+    nafe_weight : `~float`
+        Typical values for nafe_weight lie in the interval [0.05, 0.3].
+        Controls the ratio of the nafe component in the result image.
+
+    roi_p : `~tuple`
+        (row, col) of ROI top left corner pixel.
```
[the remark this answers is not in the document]
MinaSGorgy: Not really accustomed to maps so didn't think of this. If we change to taking maps instead, I will have to change this. Also, right bottom corner = (roi_p[0] + roi_f[0], roi_p[1] + roi_f[1]). Checks are applied.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+
+    roi_f : `~tuple`
+        (width, height) of ROI frame.
```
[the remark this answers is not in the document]
MinaSGorgy: Starts from roi_p.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+
+    nproc : `~int`
+        Number of processes for multiprocessing. Defaults to number of
+        available cpu cores.
+
+    hbins : `~int`
+        Number of histogram bins.
```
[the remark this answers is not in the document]
MinaSGorgy: NAFE in its core uses this histogram. No default mentioned in paper although it can be assumed to be 2^16 - 1 since this is the number of levels the paper uses. I thought if we are going to support images with even more levels to expose it as an argument. Currently default is calculated by: Freedman–Diaconis rule but I am sure something better can be done here.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+
+    nrs : `~float`
+        Sigma of artificial gaussian noise added to the input image.
+
+    n : `~int`
+        Neighborhood width used as kernel size for nafe transform. Must be an
+        odd integer.
```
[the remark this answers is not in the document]
MinaSGorgy: Whole width. The naming "n" is adopted from the paper. Can change description maybe?
The default value taken from the paper/original software.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+        Fuzzy neighborhood constants. Defaults to 12 layers each of
+        coefficient 1.
+
+    Returns
+    -------
+    enhanced_image : `~numpy.ndarray`
+        The enhanced image resulting from the linear combination of two images:
+        gamma_transform(input_image) and nafe_transform(input_image) such that:
+            enhanced_image = (1 - nafe_weight) * gamma_transform(input_image)
+                             + nafe_weight * nafe_transform(input_image)
```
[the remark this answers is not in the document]
MinaSGorgy: Will apply if once we decide on input type.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+        Sigma of artificial gaussian noise added to the input image.
+
+    n : `~int`
+        Neighborhood width used as kernel size for nafe transform. Must be an
+        odd integer.
+
+    c : `~list`
+        Fuzzy neighborhood constants. Defaults to 12 layers each of
+        coefficient 1.
+
```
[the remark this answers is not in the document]
MinaSGorgy: My bad. Will be added.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    Raises
+    ------
+    ValueError:
+        When neighborhood_width is not an odd integer or gamma is not positive.
+    """
+    if imargin is None:
+        # TODO: implement better default contrast utilizing gb, lb, gw and lw
+        # while taking in consideration roi_p and roi_f
+        imargin = (input_map.data.min(), input_map.data.max())
```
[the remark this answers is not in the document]
MinaSGorgy: Ideally according to paper/original software, no constrains applied.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    else:
+        input_map.data[input_map.data < imargin[0]] = 0
+        input_map.data[input_map.data > imargin[1]] = 0
+
+    if gamma is None:
+        gamma = _PREDEF_PARAMS.get(input_map.meta["wavelnth"], _PREDEF_PARAMS["def"])[0]
+    elif gamma < 0.0:
```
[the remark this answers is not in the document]
MinaSGorgy: Just a habit since I cam from a C background.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+        raise ValueError("gamma should be a non-negative real number! "
+                         "gamma given is:", gamma)
+
+    if omargin is None:
+        omargin = (0, 2 ** 16 - 1)
+
+    if not nafe or nafe_weight is not None and nafe_weight == 0:
+        if nafe_weight is not None and nafe_weight == 0:
+            print("Warning: setting nafe_weight to zero typically disables"
```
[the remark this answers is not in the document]
MinaSGorgy: Will apply.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+                  "nafe! nafe_weight given is:", nafe_weight)
```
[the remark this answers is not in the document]
MinaSGorgy: Will apply.
Indeed it is useless, will remove printing value.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+        enhanced_image = _transform(old_margin=imargin, new_margin=omargin, old_value=input_map.data, power=gamma)
```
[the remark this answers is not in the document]
MinaSGorgy: Indeed this is inconsistent. Will change.

sunkit_image/nafe.py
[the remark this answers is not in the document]
MinaSGorgy: No, it means my English needs improvement :sweat_smile: I meant it's equivalent and I actually don't call nafe part at all if weight is 0.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+    else:
+        if nafe_weight is None:
+            nafe_weight = _PREDEF_PARAMS.get(input_map.meta["wavelnth"], _PREDEF_PARAMS["def"])[1]
+        elif nafe_weight < 0 or nafe_weight > 1:
+            raise ValueError("nafe_weight should be between 0 and 1. "
+                             "nafe_weight given is:", nafe_weight)
+        elif nafe_weight < 0.05 or nafe_weight > 0.3:
+            print("Warning: typical values for nafe_weight should be"
+                  "between 0.05 and 0.3! nafe_weight given is:",
+                  nafe_weight)
```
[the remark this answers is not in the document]
MinaSGorgy: I thought this would be useful for debugging if the user uses sunkit-image as part of a bigger project and doesn't actually knows nafe_weight beforehand so it helps him trace things.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+                raise ValueError("roi_p should be a non-negative"
+                                 "integer numbers! roi_p given is:",
+                                 roi_p)
+            elif not isinstance(roi_f[0], int) or not isinstance(roi_f[1], int):
+                raise ValueError("roi_f should be a non-negative "
+                                 "integer numbers! roi_f given is:",
+                                 roi_f)
+            elif roi_p[0] + roi_f[0] > input_map.data.shape[0] or roi_p[1] + roi_f[1] > input_map.data.shape[1]:
+                print("Warning: part of frame is outside of given "
+                      "input_map! Clipping it.")
```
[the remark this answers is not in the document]
MinaSGorgy: I thought it maybe a small hack if the user doesn't actually know the size of the image but only wants to start from a certain point till the end, he would just enter a big number. Could be done in a cleaner way tho.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+                if roi_p[0] + roi_f[0] > input_map.data.shape[0]:
+                    roi_f[0] -= input_map.data.shape[0] - (roi_p[0] + roi_f[0])
+                if roi_p[1] + roi_f[1] > input_map.data.shape[1]:
+                    roi_f[1] -= input_map.data.shape[1] - (roi_p[1] + roi_f[1])
+                print("New frame size is:", roi_f)
```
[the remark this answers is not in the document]
MinaSGorgy: Will change.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+            map_data = input_map.data[roi_p[0]: roi_p[0] + roi_f[0], roi_p[1]: roi_p[1] + roi_f[1]]
```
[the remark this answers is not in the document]
MinaSGorgy: phew, Thanks! :smile:

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+            map_data = input_map.data
+
+        if nproc is None:
+            nproc = multiprocessing.cpu_count()
+        elif not isinstance(nproc, int) or nproc <= 0:
+            raise ValueError("nproc should be a non-negative integer number! "
+                             "nproc given is:", nproc)
+        elif nproc > multiprocessing.cpu_count():
+            print("Warning: nproc given is too big! Suggested value is:",
+                  multiprocessing.cpu_count())
```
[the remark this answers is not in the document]
MinaSGorgy: Was going for error actually but I thought maybe running 5 processes on 4 cores doesn't deserve an error. Will apply tho.

sunkit_image/nafe.py
```diff
@@ -0,0 +1,411 @@
...
+
+        if hbins is None:
+            # Freedman–Diaconis rule to calculate optimum number of
+            # historgam bins
```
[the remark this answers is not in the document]
MinaSGorgy: Should be mentioned indeed and actually will investigate this further and find a better way.

MinaSGorgy:
Thanks @dpshelio for your review :smile: I think we mainly need to decide on using a map or numpy array as input. Meanwhile I am working on testing. Will reopen a PR once all changes are applied and testing is complete.

dpshelio:
No need to re-open the PR, you can add to this, bit by bit. I've just labelled as WIP as you did on the title.

MinaSGorgy:
Sorry for late reply, last two weeks have been TOO hectic.

Applied all changes requested mainly:
- Removed NAFE class and improved multiprocessing.
- Changed to accepting a sunpy.map.Map and returning same.*
- Removed custom sub-mapping.
- Changed printing warnings to Python warn.
- Added missing documentation.
- Added testing.**

* Ideally nafe should support ROI but it's removed for now since some information is currently missing.
** Breaks CI for introducing PyTest. Missing main test comparing module output to original software because current AIA maps in samples are rejected by original software(says it treats them as colored images which aren't supported) and I am facing trouble getting AIA images due to some internal server errors.

TODO:
- Add missing test which mainly would be a very small portion of an AIA map(something like 300x300) to run fast and just test calculations.
- Support ROI (future work)

Finally for comparing results, the only maps I currently have that are accepted by the original software are 4k*4k which would take quite some time to process. Cropping them with SunPy produces rejected maps for some reason. 

@nabobalis  @dpshelio

nabobalis:
Thanks for the update. I will have to review it in more detail later on. 

Just a few comments tho:

I don't see why adding pytest would break any of the CI. 

I think ROI would be nice but not really important. People should probably pass in submaps if they want a smaller area. 

I did try to load in some data I downloaded and NAFE would not accept it, I don't know why tho.

Can you not crop a region within the NAFE windows program? Should hopefully crop the 4k image down a bit. 

I have to reiterate tho, if the before and after images are the same after we apply some normalization, has the code worked correctly?

MinaSGorgy:
Welcome @nabobalis . Take your time.

Regarding CI break: I thought I saw something suggesting that in the error log but seems there isn't. Will investigate this. Any hints though?

Regarding the ROI: the thing is the original software most probably doesn't discard the out of ROI pixels. ie it doesn't really crop the image but just calculates new values for the ROI pixels including in the calculations the out of ROI neighbor pixels. It even does this with a ratio which we can change in the ini file but I am not 100% sure it's linear. Hence only full size images would provide reasonable comparison.

Regarding before and after images: if I am getting it correctly that you are even questioning the benefit of the filter itself against simpler normalization filters, this I can test using original software.

nabobalis:
I'm not sure what to hint about, the CI has all passed bar the install one. That is a known fault and I need to fix it separately. 

Yes I guess I'm questing the need for this filter if it's just the same as normalizing it with a simple function.  Which makes me suspect that if the original method is "correct", then there is a mistake somewhere in the version here.

PR #28, opened 2019-05-24
if not dev, dont build changelog

nabobalis:
Update for changelog from sunpy master.


Cadair:
We should also pull this into ndcube. Also we should get the template to the point where we can retemplate things for these kind of changes.

nabobalis:
IF ONLY SOMEONE WOULD MERGE IN MY PR FOR NDCUBE

nabobalis:
IF ONLY MY PR COULD BE MERGED IN SO I COULD LEAVE THE ROOM

nabobalis:
IM NOT MAD IM JUST ANGRY

PR #30, opened 2019-05-28
Implemented Multi-scale Gaussian Normalization

### LINKED ISSUE #1: Multi-scale Gaussian Normalisation
The original attempt is located here (https://github.com/sunpy/sunpy/pull/1899).
Please note there are is at least one issue with that code (https://github.com/sunpy/sunpy/pull/1899#discussion_r104646604).

This needs to be moved into this module and have the following done:

### Base
- [ ] Find the original AIA image used in the [original paper](https://link.springer.com/article/10.1007/s11207-014-0523-9). Note that the paper gets the time of the image wrong.
- [ ] Test to compare to known IDL output ([IDL version](http://eagle.imaps.aber.ac.uk/mgn.pro))
- [ ] We need to find the difference between this code, the IDL code and the [original paper](https://link.springer.com/article/10.1007/s11207-014-0523-9). There are known inconsistencies that need to be tracked down and figured out. Once we know these, we can email the original author to ask. 
- [ ] Compared to another version in [Python.](https://git.ias.u-psud.fr/ebuchlin/aia-movie/blob/master/medocimage/mgn.py) Maybe both can exist within this package.

### Extra
- [ ] Optimized as much as possible for memory and CPU time.
- [ ] 100% test coverage
- [ ] Documentation and a worked example.

vatch123:
### Description

* Contains both @Cadair and @ebuchlin code. As both the implementations are accurate with minor differences and both provide visually acceptable outputs.
* Cadair's code has been modified as mentioned in this [comment](https://github.com/sunpy/sunkit-image/issues/1#issuecomment-472108908).
* Though Eric's code has minor differences from the paper as pointed out by him [here](https://github.com/sunpy/sunkit-image/issues/1#issuecomment-472184298).
* Have a look at this [gist](https://gist.github.com/vatch123/2a77bb68c414463d7c6efdedc59b121a) to see the differences in the output of both.

TODO:
- [x] Review of PR
- [x] ~100% test line coverage 
- [x] Code formatted/linted.
- [x] Example and documentation is detailed.

Fixes #1


nabobalis:
I think the main point is that we don't want to have both versions in our code but one unified version which we can switch to (maybe with a keyword) if we want.

vatch123:
> I think the main point is that we don't want to have both versions in our code but one unified version which we can switch to (maybe with a keyword) if we want.

Then it is better that we choose one because both are essentially the same. The only difference is that one has some weights and the other doesn't (which is equivalent to setting the weights 1).

nabobalis:
The output doesn't look similar. So what is causing that?

vatch123:
> The output doesn't look similar. So what is causing that? 

* The image is normalized before being processed which is then used later as mentioned in this [comment](https://github.com/sunpy/sunkit-image/issues/1#issuecomment-472184298).
* The weights **g<sub>i</sub>** as mentioned in the paper have not been used.
* The calculated values after applying the transform has been again brought between (0,1) from (-pi/2, pi/2). This is nowhere mentioned in the paper. 

Please have a look at this [comment](https://github.com/sunpy/sunkit-image/issues/1#issuecomment-472184298) for greater details as how the code varies from the paper.


> I think the main point is that we don't want to have both versions in our code but one unified version which we can switch to (maybe with a keyword) if we want.

Then working on Cadair's code is better because now it is exactly similar to the paper and gives better visual output too.

sunkit_image/enhance.py
```diff
@@ -0,0 +1,177 @@
...
+
+    """
+    Multi-scale Gaussian normalization.
+
+    Parameters
+    ----------
+    data : `numpy.ndarray`
+        Image to be transformed.
+    sigma : `list`, optional
+        Range of guassian widths to transform over.
```
ebuchlin:
```suggestion
        Range of Gaussian widths to transform over.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,177 @@
...
+    k : `float`, optional
+        Controls the severity of the arctan transformation.
+        Defaults to 0.7
+    gamma : `float`, optional
+        The value used to calulcate the  global gamma-transformed image.
```
ebuchlin:
```suggestion
        The value used to calculate the  global gamma-transformed image.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,177 @@
...
+        Ideally should be between 2.5 to 4.
+        Defaults to 3.2
+    h : `float`, optional
+        Weight of global filter to gaussian filters.
```
ebuchlin:
```suggestion
        Weight of global filter to Gaussian filters.
```

sunkit_image/enhance.py
ebuchlin: Also: specify what is the default?

sunkit_image/enhance.py
```diff
@@ -0,0 +1,177 @@
...
+
+    # This is rectified version of Stuart's Code.
+    if weights is None:
+        weights = np.ones(len(sigma))
+
+    # 1. Replace spurious negative pixels with zero
+    data[data <= 0] = 1e-15  # Makes sure that all values are above zero
+    image = np.empty(data.shape, dtype=data.dtype)
+    conv = np.empty(data.shape, dtype=data.dtype)
+    sigmaw = np.empty(data.shape, dtype=data.dtype)
```
ebuchlin: Use `np.empty_like(data)`?

ebuchlin:
About visual output comparison (as seen on the [gist](https://gist.github.com/vatch123/2a77bb68c414463d7c6efdedc59b121a)): the output depends on the input parameters, which are not specified in the gist. I guess that they are the default parameters, but the default parameters are not the same for `mgn1()` and `mgn2()` in `enhance.py`.

This can be seen immediately for `sigma`, but it is also the case for the global weights: the defaults `h=0.7` for `mgn1()` and `weight=0.3` for `mgn2()` are **not** equivalent (although `0.3 == 1 - 0.7`), because of the arctan normalization from [-π/2, π/2] to [0,1] in the end of  `mgn2()`. I would say (untested) that the equivalency would be given by `1 - h == weight / pi`, that is `weight = .3 * pi` (about .94) for `h = 0.7`. The fact that .94 is much closer to 1 than 0.3 would explain why the (gamma-corrected) background image dominates in both "Eric's Outputs" in @vatch123's gist, while the pure Gaussian normalization dominates in both "Cadair's Outputs"  in the gist.

IMHO the definition of "weight" (i.e. with arctan normalization to [0,1]) is much more meaningful that that of "h".
This is almost independent of the implementation itself (I think both are fine, and I appreciate the speed and memory optimizations made for `mgn1()`).

vatch123:
Yes, the default parameters in both the implementation are different.  But after discussion with @nabobalis and @wafels we have decided to only have ```mgn1()``` just because it is an exact replica of the paper and the parameters also follow the paper.

sunkit_image/enhance.py
```diff
@@ -0,0 +1,177 @@
...
+    weights=None,
+    truncate=3
+):
+
+    """
+    Multi-scale Gaussian normalization.
+
+    Parameters
+    ----------
+    data : `numpy.ndarray`
```
vatch123: Wouldn't it be better that both input and output are ```sunpy.map``` rather than ```numpy.ndarray```

nabobalis: I'm not sure we need to limit it to just a SunPy map. If the code works on a standard array, I think we should keep it that way.

vatch123: Okay

wafels: I agree with nabobalis here.  The base algorithms should accept numpy arrays in general.  Later we can develop higher level functions that accept SunPy maps as arguments.

vatch123: Yes I am keeping them as numpy arrays

nabobalis: We do want to support sunpy maps as well.

wafels: use astropy units where appropriate.  For example, the sigma argument here should be `sigma=[1.25, 2.5, 5, 10, 20, 40]*u.pix`

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+        Image to be transformed.
+    sigma : `list`, optional
+        Range of Gaussian widths to transform over.
+        Defaults to ``[1.25, 2.5, 5, 10, 20, 40]``.
+    k : `float`, optional
+        Controls the severity of the arctan transformation.
```
wafels: What does "severity" mean?

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+        Ideally should be between 2.5 to 4.
+        Defaults to 3.2
+    h : `float`, optional
+        Weight of global filter to Gaussian filters.
+        Defaults to 0.7
+    weights : `list`, optional
+        Used to weight all the transformed images during the calculation of the
+        final image. If not specificed, all weights are one.
+    width : `int`
+        An odd integer defining the width of the kernel to be convolved.
```
wafels: what is width?  It is not in the function signature.

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+    truncate : `int`
+        The number of sigmas to truncate the kernel.
```
wafels: This needs a little bit more explanation.  Is this full width?

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+
+    .. notes::
+        In practice, the weights and h may be adjusted according to the desired
+        output, and also according to the type of input image
+        (e.g. wavelength or channel).
+        For most purposes, the weights can be set
+        equal for all scales.
+    """
+
+    # This is rectified version of Stuart's Code.
```
wafels: We don't need this comment - it won't make sense in a few years from now.

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+    data[data <= 0] = 1e-15  # Makes sure that all values are above zero
+    image = np.zeros(data.shape, dtype=data.dtype)
+    conv = np.zeros(data.shape, dtype=data.dtype)
+    sigmaw = np.zeros(data.shape, dtype=data.dtype)
+
+    for s, weight in zip(sigma, weights):
+        # 2 & 3 Create kernel and convolve with image
+        ndimage.filters.gaussian_filter(data, sigma=s,
+                                        truncate=truncate, mode='nearest', output=conv)
+        # 5. Calculate difference between image and the local mean image,
```
nabobalis: Is this meant to be step 4?

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # This is rectified version of Stuart's Code.
+    if weights is None:
+        weights = np.ones(len(sigma))
+
+    # 1. Replace spurious negative pixels with zero
+    data[data <= 0] = 1e-15  # Makes sure that all values are above zero
+    image = np.zeros(data.shape, dtype=data.dtype)
+    conv = np.zeros(data.shape, dtype=data.dtype)
+    sigmaw = np.zeros(data.shape, dtype=data.dtype)
```
wafels: `image = np.zeros_like(data)` and also for the rest.

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+        # 5. Calculate difference between image and the local mean image,
+        # square the difference, and convolve with kernel. Square-root the
+        # resulting image to give ‘local standard deviation’ image sigmaw
+        conv = data - conv
+        ndimage.filters.gaussian_filter(conv ** 2, sigma=s,
+                                        truncate=truncate, mode='nearest', output=sigmaw)
+        np.sqrt(sigmaw, out=sigmaw)
+        conv /= sigmaw
+
+        # 6. Apply arctan transformation on Ci to give C'i
```
wafels: Refer to equations in the paper when appropriate.

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+        Image to be transformed.
+    sigma : `list`, optional
+        Range of Gaussian widths to transform over.
+        Defaults to ``[1.25, 2.5, 5, 10, 20, 40]``.
+    k : `float`, optional
+        Controls the severity of the arctan transformation.
+        Defaults to 0.7
+    gamma : `float`, optional
+        The value used to calculate the  global gamma-transformed image.
+        Ideally should be between 2.5 to 4.
```
wafels: Where do these numbers come from?

changelog/30.feature.rst
```diff
@@ -0,0 +1 @@
+Added a function (`sunkit_image.enhance.mgn`) for applying Multi-scale Gaussian Normalization to a image.
```
nabobalis:
```suggestion
Added a function (`sunkit_image.enhance.mgn`) for applying Multi-scale Gaussian Normalization to an image (`numy.array`).
```

docs/code_ref/enhance.rst
```diff
@@ -0,0 +1,7 @@
+*******
+enhance
+*******
+
+This package implements enhancement routines for solar physics data.
```
nabobalis: I am not sure I like calling these enhancement routines.

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
+"""
+==================================
+Multi-scale Gaussian Normalization
+==================================
+This example applies Multi-scale Gaussian Normalization to a sunpy map.
```
nabobalis:
```suggestion

This example applies Multi-scale Gaussian Normalization to a sunpy map.
```

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
...
+The example uses `sunkit_image.enhance` to apply the filter.
```
nabobalis:
```suggestio
```

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+# Start by importing the necessary modules.
```
nabobalis:
```suggestion
```

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
...
+import matplotlib.pyplot as plt
+
+import sunpy.map
+import sunpy.data.sample
+
+import sunkit_image.enhance as enhance
+
+###########################################################################
+# Sunpy sample data contains a number of suitable maps, where the sunpy.data.sample.NAME
```
nabobalis:
```suggestion
# SunPy sample data contains a number of suitable images, which we will use here.
```

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
...
+# returns the location of the given FITS file.
```
nabobalis:
```suggestion
```

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
...
+
+aia_map = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+
+###########################################################################
+# Applying both the variants of Multi-scale Gaussian Normalization
```
nabobalis: Both?

I would also like some explanation as to what it is and why you might want to use it. Think of these as mini tutorials.

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
...
+
+out = enhance.mgn(aia_map.data)
+out = sunpy.map.Map(out, aia_map.meta)
```
nabobalis: Explain this step in the example.

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
...
+
+###########################################################################
+# The resulting sunpy.map are plotted
```
nabobalis:
```suggestion
# The resulting Map is plotted.
```

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
...
+
+fig = plt.figure()
```
nabobalis:
```suggestion
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
+"""
+This module contains functions that can be used to enhance the entire solar image.
```
nabobalis:
```suggestion
This module contains functions that will enhance the entire image.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+        Defaults to 3.2
+    h : `float`, optional
+        Weight of global filter to Gaussian filters.
+        Defaults to 0.7
+    weights : `list`, optional
+        Used to weight all the transformed images during the calculation of the
+        final image. If not specificed, all weights are one.
+    width : `int`
+        An odd integer defining the width of the kernel to be convolved.
+    truncate : `int`
```
nabobalis:
```suggestion
    truncate : `int`, optional 
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+        The number of sigmas to truncate the kernel.
+        Defaults to 3
```
nabobalis:
```suggestion

```

sunkit_image/enhance.py
nabobalis:
```suggestion
        The number of sigmas (defaults to 3) to truncate the kernel.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Multi-scale Gaussian normalization.
+
+    Parameters
+    ----------
+    data : `numpy.ndarray`
+        Image to be transformed.
+    sigma : `list`, optional
+        Range of Gaussian widths to transform over.
+        Defaults to ``[1.25, 2.5, 5, 10, 20, 40]``.
```
nabobalis: use astropy units where appropriate.  For example, the sigma argument here should be `sigma=[1.25, 2.5, 5, 10, 20, 40]*u.pix`

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+        final image. If not specificed, all weights are one.
+    width : `int`
+        An odd integer defining the width of the kernel to be convolved.
+    truncate : `int`
+        The number of sigmas to truncate the kernel.
+        Defaults to 3
+
+    Returns
+    -------
+    image: `numpy.ndarray`
```
nabobalis:
```suggestion
    `numpy.ndarray`
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,122 @@
...
+        Normalized image.
+
+    Reference
+    ---------
+    * Morgan, Huw, and Miloslav Druckmuller.
+    "Multi-scale Gaussian normalization for solar image processing."
+    arXiv preprint arXiv:1403.6613 (2014).
+    Ref: Sol Phys (2014) 289: 2945. doi:10.1007/s11207-014-0523-9
+
+    .. notes::
```
nabobalis: This should be before the parameters section.

sunkit_image/tests/test_enhance.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@figure_test
+@pytest.mark.remote_data
+def test_mgn(smap):
+
+    out = enhance.mgn(smap.data)
+    out = sunpy.map.Map(out, smap.meta)
+
+    fig = plt.figure()
```
nabobalis:
```suggestion
```

sunkit_image/tests/test_enhance.py
```diff
@@ -0,0 +1,45 @@
...
+    A = np.array([[1., 1., 1., 1.],
+                 [1., 5., 5., 1.],
+                 [1., 5., 5., 1.],
+                 [1., 1., 1., 1.]])
+    return A
+
+
+def test_multiscale_gaussian(map_test):
+
+    # Assuming the algorithm works fine then the below two should be equal.
```
nabobalis: Why would these two be equal?

sunkit_image/tests/test_enhance.py
```diff
@@ -0,0 +1,45 @@
...
+    expect1 = enhance.mgn(map_test, [1])
+    expect2 = enhance.mgn(map_test, [1, 1])
+
+    assert np.allclose(expect1, expect2)
```
nabobalis: I do wonder if this test is useful. What is the output from the tests if you use the maps you created for the fourier tests?

sunkit_image/tests/test_enhance.py
```diff
@@ -0,0 +1,45 @@
...
+import sunpy.data.sample
+from sunpy.tests.helpers import figure_test
+
+import sunkit_image.enhance as enhance
+
+
+@pytest.fixture
+@pytest.mark.remote_data
+def smap():
+    return sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
```
nabobalis: Switch to the sunpy test data for the figures. It would allow us to drop the remote_data flag.

nabobalis:
I think going forward (for all PRs) we do want to have more comments in the code that detail any parts of the code that might be hard to understand and to refer to what equations or parts of the paper a section of code is implementing.

sunkit_image/enhance.py
ebuchlin: If the image is simply a `ndarray` and not a `Map`, I am not sure that I understand why there should be units here... This would be the only astropy dependence of the function.

examples/enhance/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,33 @@
+"""
```
nabobalis: Move the example file one folder higher.
For now just put them in `examples/` and no subfolders.

sunkit_image/enhance.py
vatch123: These are the values which were recommended in the paper

docs/code_ref/enhance.rst
vatch123: Then what should we call them as they work for the entire image

sunkit_image/enhance.py
vatch123: Yes for a simple `ndarray` why we need units

sunkit_image/enhance.py
vatch123: Yes for sure

sunkit_image/enhance.py
vatch123: It is the recommended value in the paper

sunkit_image/tests/test_enhance.py
vatch123: Since both the kernel sigmas are equal in the second case and for all the images after convolving with kernels, weighted mean is calculated so the final output of first and second will be same.

sunkit_image/tests/test_enhance.py
vatch123: I tried with them first, the values which the code output is very difficult to verify manually

docs/code_ref/enhance.rst
nabobalis: Good question, I guess we can leave this for now until I come up with a better name.

sunkit_image/enhance.py
nabobalis: Fair point.

sunkit_image/enhance.py
nabobalis: We can mention that.

sunkit_image/tests/test_enhance.py
nabobalis: But if we don't plan to verify this manually, we can expand the range of tests.

vatch123: What do you mean by range?

sunkit_image/enhance.py
vatch123: Removed it was an error.

sunkit_image/enhance.py
vatch123: I did some research and feel that it is the half width maximum.
Screenshot from 2019-06-08 23-38-14
This is place in the paper where they have mentioned it. Maybe you can make it a little more clear.

sunkit_image/tests/test_enhance.py
nabobalis: I think one test for such a complex function feels too little.

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+
+###########################################################################
+
+# SunPy sample data contains a number of suitable images, which we will use here.
+aia_map = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+
+###########################################################################
+
+# Applying Multi-scale Gaussian Normalization on a solar image.
+# The ``mgn`` function takes a ``numpy.ndarray`` as a input so we will pass only
```
nabobalis:
```suggestion
# The `sunkit_image.enhance.mgn` function takes a `numpy.ndarray` as a input so we will pass only
```

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+# the data part of ``sunpy.map.Map``
```
nabobalis:
```suggestion
# the data part of `~sunpy.map.GenericMap`
```

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+out = enhance.mgn(aia_map.data)
+
+# The value returned is also a ``numpy.ndarray`` so we convert it back to
+# ``sunpy.map.Map``
+out = sunpy.map.Map(out, aia_map.meta)
+
+###########################################################################
+
```
nabobalis:
```suggestion
```

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+# The resulting Map is plotted.
```
nabobalis:
```suggestion
# The resulting map is plotted.
```

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+aia_map = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+
+###########################################################################
+
+# Applying Multi-scale Gaussian Normalization on a solar image.
+# The ``mgn`` function takes a ``numpy.ndarray`` as a input so we will pass only
+# the data part of ``sunpy.map.Map``
+out = enhance.mgn(aia_map.data)
+
+# The value returned is also a ``numpy.ndarray`` so we convert it back to
```
nabobalis:
```suggestion
# The value returned is also a numpy.ndarray so we convert it back to
```

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+# ``sunpy.map.Map``
```
nabobalis:
```suggestion
# a  sunpy.map.Map
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+def mgn(
+    data,
+    sigma=[1.25, 2.5, 5, 10, 20, 40],
+    k=0.7,
+    gamma=3.2,
+    h=0.7,
+    weights=None,
+    truncate=3
+):
+
```
nabobalis:
```suggestion
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+    Extreme ultra-violet images of the corona contain information over a wide range of spatial scales,
+    and different structures such as active regions, quiet Sun, and filament channels contain information
+    at very different brightness regimes. The MGN method normalises an image by using the local mean and
+    standard deviation calculated using a Gaussian-weighted sample of local pixels. This normalised image
+    is transformed by the arctan function (similar to a gamma transformation). This is applied over
+    several spatial scales, and the final image is a weighted combination of the normalised components.
+    The method reveals information at the finest scales whilst maintaining enough of the larger-scale
+    information to provide context. It also intrinsically flattens noisy regions and can reveal structure
+    in off-limb regions out to the edge of the field of view.
```
nabobalis: Is this just not the abstract of the paper? We can't just copy and paste it.

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+
+    .. note::
+        In practice, the weights and h may be adjusted according to the desired output, and also according
+        to the type of input image
+        (e.g. wavelength or channel).
+        For most purposes, the weights can be set
+        equal for all scales.
```
nabobalis: This block seems very broken up?

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+    ----------
+    data : `numpy.ndarray`
+        Image to be transformed.
+    sigma : `list`, optional
+        Range of Gaussian widths (i.e. the standard deviation of the Gaussian kernel) to transform over.
+        Defaults to ``[1.25, 2.5, 5, 10, 20, 40]``.
+    k : `float`, optional
+        Controls the severity of the arctan transformation. The scaling factor multiplied with each
+        Gaussian transformed image before applying the arctan transformation.
+        Defaults to 0.7
```
nabobalis:
```suggestion
        Defaults to 0.7.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+    gamma : `float`, optional
+        The value used to calculate the  global gamma-transformed image.
```
nabobalis:
```suggestion
        The value used to calculate the global gamma-transformed image.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+        Ideally should be between 2.5 to 4 according to the paper.
+        Defaults to 3.2
```
nabobalis:
```suggestion
        Defaults to 3.2.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+    h : `float`, optional
+        Weight of global filter to Gaussian filters.
+        Defaults to 0.7
```
nabobalis:
```suggestion
        Defaults to 0.7.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+    weights : `list`, optional
+        Used to weight all the transformed images during the calculation of the
+        final image. If not specificed, all weights are one.
```
nabobalis:
```suggestion
        final image. If not specified, all weights are one.
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+
+        image += conv
+
+    # delete these arrays here as it reduces the total memory consumption when
+    # we create the Cprime_g temp array below.
+    del conv
+    del sigmaw
+
+    # 8. Take weighted mean of C'i to give a weighted mean locally normalised
+    # image.
```
nabobalis:
```suggestion
```

sunkit_image/enhance.py
```diff
@@ -0,0 +1,140 @@
...
+        conv *= weight
+
+        image += conv
+
+    # delete these arrays here as it reduces the total memory consumption when
+    # we create the Cprime_g temp array below.
+    del conv
+    del sigmaw
+
+    # 8. Take weighted mean of C'i to give a weighted mean locally normalised
```
nabobalis:
```suggestion
    # 8. Take weighted mean of C'i to give a weighted mean locally normalised image.
```

nabobalis:
@ebuchlin Would it be possible for you to do another review for us?

nabobalis:
Could you run black, isort and docformater on your PR?

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+
+import sunpy.map
+import sunpy.data.sample
+
+import sunkit_image.enhance as enhance
+
+###########################################################################
+
+# SunPy sample data contains a number of suitable images, which we will use here.
+aia_map = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
```
nabobalis: We should plot the original map as well to showcase the difference.

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+import matplotlib.pyplot as plt
+
+import sunpy.map
+import sunpy.data.sample
+
+import sunkit_image.enhance as enhance
+
+###########################################################################
+
```
nabobalis:
```suggestion
```

examples/multiscale_gaussian_normalization.py
```diff
@@ -0,0 +1,36 @@
...
+# SunPy sample data contains a number of suitable images, which we will use here.
+aia_map = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+
+###########################################################################
+
```
nabobalis:
```suggestion
```

vatch123:
This is done too, I guess

nabobalis:
Did you run the linter/formaters on the code? 

I think it is ready but I would like either @ebuchlin or @Cadair to review it once more (before I merge this) since they helped here.

Until that happens, we should go back to working on the OCCULT-2 code!

vatch123:
Okay, I will begin on it again from tomorrow.
Yes I ran the linters/formaters

nabobalis:
Thanks for the review @ebuchlin! If you want to run the code, the most straightforward way would be to clone the branch this code is on by doing `git clone -b mgn https://github.com/vatch123/sunkit-image.git` then either installing the package or I think if the deps are just installed, you should be able to run the file (maybe?).

ebuchlin:
> Thanks for the review @ebuchlin! If you want to run the code, the most straightforward way would be to clone the branch this code is on by doing `git clone -b mgn https://github.com/vatch123/sunkit-image.git` then either installing the package or I think if the deps are just installed, you should be able to run the file (maybe?).

Thanks, I could just run the example file, and the output was as expected.

sunkit_image/enhance.py
```diff
@@ -0,0 +1,125 @@
...
+      "Multi-scale Gaussian normalization for solar image processing."
+      arXiv preprint arXiv:1403.6613 (2014).
+      Ref: Sol Phys (2014) 289: 2945. doi:10.1007/s11207-014-0523-9
+    """
+
+    if weights is None:
+        weights = np.ones(len(sigma))
+
+    # 1. Replace spurious negative pixels with zero
+    data[data <= 0] = 1e-15  # Makes sure that all values are above zero
```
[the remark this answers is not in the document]
nabobalis: If we set a keyword that makes these stage optional? Would that be an ok compromise?

nabobalis:
Thanks @vatch123 and @ebuchlin for your reviews!

PR #22, opened 2019-04-05
Declassed NoiseLevelEstimation

m-squared96:
### Description

Fixes #18 
Separates NoiseLevelEstimation into function-equivalents. Direct conversion, no new functionality added or old functionality lost.

PR #38, opened 2019-08-20
Noise

### LINKED ISSUE #18: De-Class NoiseLevelEstimation

vatch123:
### Description

Continuation of the PR #22. Fixed the variable names as mentioned in the comments on that thread. No implementation related changes have been made apart from the fact that the `test_noise.py` now uses pytest. The comments which required knowledge of implementation have been left as it is.

Fixes #18


sunkit_image/utils/noise.py
```diff
@@ -1,270 +1,309 @@
...
+    confidence : `float`, optional
+        Confidence interval to determine the threshold for the weak texture.
+        In this algorithm, this value is usually set the value very close to one.
+        Defaults to 0.99.
+    iterations : `int`, optional
+        Number of iterations,  defaults to 3.
+
+    Returns
+    ----------
+    nlevel: `numpy.ndarray`
```
nabobalis: This is all in the dict now, so this has to be reflected here.

PR #49, opened 2020-05-02
updating to use sunpy.map.Map

hayesla:
### Description
I've updated the example of `sunkit_image.trace` to use `sunpy.map` and convert the pixel coordinates to world coordinates in the plotting


nabobalis:
Thanks @hayesla.

PR #59, opened 2021-04-05
Increasing Test coverage

### LINKED ISSUE #33: Increase test coverage sunkit-image
We need to be aware of our test coverage and with #40 it will drop.

So rising this back as high as possible is important.

Missing direct tests:
- `fit_polynomial_to_log_radial_intensity`
- `normalize_fit_radial_intensity`
- `calculate_fit_radial_intensity`
- `intensity_enhance`
- `calc_gamma`
- `remove_duplicate`
- `points_in_poly`
- `reform2d`
- `Lamb_Oseen`

jeffreypaul15:
### Description

Fixes #33 
Tests have yet to be added.
- [x] fit_polynomial_to_log_radial_intensity
- [x] normalize_fit_radial_intensity
- [x] calculate_fit_radial_intensity
- [x] intensity_enhance
- [x] calc_gamma
- [x] remove_duplicate
- [x] points_in_poly
- [x] reform2d
- [x] Lamb_Oseen


jeffreypaul15:
@nabobalis Are these tests fine or should I have to manually store the lists and then check for `asstert np.allclose` instead of calculating them and then checking?

nabobalis:
The tests look fine, I assume the ipython notebook was just for your use?

jeffreypaul15:
> The tests look fine, I assume the ipython notebook was just for your use?
> 
> Figure tests or normal unit test would be ok. The figure-tests is broken due to a deprecation warning, probably best to ignore it for now in setup.cfg

Oh yeah, sorry about the notebook I'll remove it

jeffreypaul15:
@nabobalis methods from `Lamb_Oseen` class are already tested but a few are missing, I shall add them slowly.
Are the remaining tests written fine?
I shall go about debugging them simultaneously.

jeffreypaul15:
py37-oldestdeps [linux] is failing and I'm a bit confused as to why.
The part where it fails runs fine on py38.

sunkit_image/utils/tests/test_utils.py
```diff
@@ -108,10 +110,84 @@ def test_find_pixel_radii(smap):
...
 
     # Test that the new scale is used
-    pixel_radii = find_pixel_radii(smap, scale=2 * smap.rsun_obs)
+    pixel_radii = utils.find_pixel_radii(smap, scale=2 * smap.rsun_obs)
     assert_quantity_allclose(np.max(pixel_radii).value, known_maximum_pixel_radius / 2)
 
 
 def test_get_radial_intensity_summary():
     # TODO: Write some tests.
     pass
```
nabobalis, line 119: What about this one?

nabobalis:
> py37-oldestdeps [linux] is failing and I'm a bit confused as to why.
> The part where it fails runs fine on py38.

Well a package difference means the test output is different.

jeffreypaul15:
> > py37-oldestdeps [linux] is failing and I'm a bit confused as to why.
> > The part where it fails runs fine on py38.
> 
> Well a package difference means the test output is different.

Yeah the output is different for both of the tests, is there a way I can get a different output for the different versions?

nabobalis:
> > > py37-oldestdeps [linux] is failing and I'm a bit confused as to why.
> > > The part where it fails runs fine on py38.
> > 
> > 
> > Well a package difference means the test output is different.
> 
> Yeah the output is different for both of the tests, is there a way I can get a different output for the different versions?

Run the oldest-deps build locally or install the package versions that have changed in the oldest-deps build.

sunkit_image/utils/tests/test_utils.py
jeffreypaul15: Oh yeah, I forgot about this. I'll add this.

nabobalis:
> > Run the oldest-deps build locally or install the package versions that have changed in the oldest-deps build.
> 
> I've checked for the version of change in `skimage` as that seemed to be causing the change in output.

It would be better to understand what is causing the change and then decide if we need to change the minimum package version.

sunkit_image/utils/tests/test_utils.py
```diff
@@ -108,10 +114,106 @@ def test_find_pixel_radii(smap):
...
+
+    map_r = utils.find_pixel_radii(smap, scale=smap.rsun_obs).to(u.R_sun)
+
+    nbins = radial_bin_edges.shape[1]
+
+    lower_edge = [map_r > radial_bin_edges[0, i].to(u.R_sun) for i in range(0, nbins)]
+    upper_edge = [map_r < radial_bin_edges[1, i].to(u.R_sun) for i in range(0, nbins)]
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=RuntimeWarning)
```
nabobalis, line 135: What warning happens?

nabobalis:
Probably worth adding a changelog considering how many tests are here.

jeffreypaul15:
> It would be better to understand what is causing the change and then decide if we need to change the minimum package version.

I suppose, I shall dig a little deeper and try to find out.

nabobalis:
> Would trivial do?

Works for me.

sunkit_image/utils/tests/test_utils.py
nabobalis: This should probably be caught with pytest.warns and matched to a message.

jeffreypaul15: Warnings arise when we take the mean of an empty slice, `summary` defaults to `np.mean` and I've left it as it is.
https://github.com/sunpy/sunkit-image/blob/a55393845f83e1a75128e567744b5a3ee3294c73/sunkit_image/utils/utils.py#L171

I've used this as a reference.

nabobalis: Fair enough.

jeffreypaul15: Should I use PyTest instead?

nabobalis: No, its fine.

jeffreypaul15:
After looking around, I've found that skimage's `measure.grid_points_in_poly` doesn't produce the right output before version 1.18. Based on the discussion at https://github.com/scikit-image/scikit-image/issues/3892, it seems to have been fixed.

Ideally `grid_points_in_poly` takes in an input of a grid (ex: 3,3) and edges of the polygon and outputs a boolean array the same shape of the grid, returning true for all the on the grid that are part-of/ inside the given polygon. 

```python
# version < 0.18.0, produces buggy output
poly = [(0, 0), (0, 1), (0, 2), (1, 2), (2, 2), (2, 0)]
mask = measure.grid_points_in_poly((3,3), poly)
print(mask)
[[ True  True False]
 [ True  True False]
 [False False False]]

# version >= 0.18.0
poly = [(0, 0), (0, 1), (0, 2), (1, 2), (2, 2), (2, 0)]
mask = measure.grid_points_in_poly((3,3), poly)
print(mask)
array([[ True,  True,  True],
       [ True,  True,  True],
       [ True,  True,  True]])
```
In the above example the points 
```python
[(0,2), (1,2), (2,0), (2,1), (2,2)]
```
all lie on the edge of the polygon but based on previous versions of `skimage` produces `False` for those points.

This was buggy before and didn't return the right boolean array but it seemed to have been fixed in 1.18. 
I suggest we change the default version because of this.

nabobalis:
Thanks for looking in to it. Change the version in setup.cfg and bump the version in tox.ini for the oldest deps.

jeffreypaul15:
> Thanks for looking in to it. Change the version in setup.cfg and bump the version in tox.ini for the oldest deps.

Will do.

nabobalis:
> > Thanks for looking in to it. Change the version in setup.cfg and bump the version in tox.ini for the oldest deps.
> 
> Will do.

I think this will want a seperate changelog as well.

jeffreypaul15:
(Core Tests py38 [linux]) seems to be failing, is it because of the change in setup or tox.ini files?

nabobalis:
Looks like some internet issues.

changelog/59.bugfix.rst
```diff
@@ -0,0 +1 @@
+Changed minimum version of `skimage` to 0.18.0, preventing faulty code in :meth:`sunkit-image.utils.points_in_poly`.
```
nabobalis, line 1:
```suggestion
Increased the minimum version of "skimage" to 0.18.0, preventing faulty code in :meth:`sunkit-image.utils.points_in_poly`.
```

tox.ini
```diff
@@ -54,7 +54,7 @@ deps =
 
     # Oldest deps we pin against.
     oldestdeps: sunpy<2.1
-    oldestdeps: scikit-image<0.17.0
+    oldestdeps: scikit-image<=0.18.0
```
nabobalis, line 57:
```suggestion
    oldestdeps: scikit-image<=0.19.0
```

jeffreypaul15: 1.19 isn't out yet though right?

nabobalis: Sure.

jeffreypaul15: I'm a little confused as to how `oldestdeps` works though. From what I understand, we test for every version up until < 0.19.0?

nabobalis: It only tests one version.

jeffreypaul15: > It only tests one version.

Which version does it test in the above example?

nabobalis: 0.18.0

jeffreypaul15:
@nabobalis I suppose all the tests are fine?

nabobalis:
Yeah the tests are fine.

jeffreypaul15:
Would we want to add `Lamb_Oseen` tests on this PR itself?

nabobalis:
> Would we want to add `Lamb_Oseen` tests in this PR itself?

Up to you.

jeffreypaul15:
> Up to you.

I'll add them here itself.

jeffreypaul15:
This should cover all the required tests for now. If all the tests pass, would this be good to merge?

nabobalis (approved):

nabobalis:
Thanks for the PR @jeffreypaul15

PR #60, opened 2021-04-21
Add gamma_min/max as kwargs

wtbarnes:
### Description

In the IDL implementation (see https://solarphysics.aber.ac.uk/mgn.pro), the minimum and maximum values for the gamma-transform, called `a0` and `a1`, are input arguments. However, in this translation, they are fixed as the min and max values of the image. This PR makes them arguments that default to the min and max of the image. This should probably include a test for non-None values of these kwargs...

### DIFF
--- a/changelog/60.feature.rst
+++ b/changelog/60.feature.rst
@@ -1,0 +1,1 @@
+The minimum and maximum values of the gamma transform can now be specified for :func:`sunkit_image.enhance.mgn`.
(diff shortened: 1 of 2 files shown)


nabobalis (approved):

nabobalis:
Thanks for the PR!. I don't think we need to worry about that test.

wtbarnes:
Oops  I just realized I don't have precommit installed on my local version of sunkit-image...

nabobalis:
It's fine, I dealt with all of that.

PR #70, opened 2021-10-29
CI Fix

nabobalis:
Aim to release an update.

PR #72, opened 2021-11-19
Dask quantity fixes

wtbarnes:
See breakage in #70.

This fixes an issue, where in newer versions of numpy, astropy quantities that were being represented as Dask arrays could not be properly indixed. Now, we don't worry about units until the end and just reattach them to the final product.

This is technically a breaking change since I've removed the `pre_check_hook` option and renamed the `post_check_hook` to `array_check`.


nabobalis (approved):

nabobalis:
Doc failure is the same error as before. The example does something that breaks it.

nabobalis (approved):

PR #93, opened 2022-09-02
Add ability to create Carrington maps

dstansby:
I started a new sub-module for reprojection stuff. I think there's a couple of other bits it would be nice to have in here, but to start with here's code to create Carrington maps.


sunkit_image/reproject.py
```diff
@@ -0,0 +1,106 @@
...
+        See the reproject docs for more info.
+
+    Returns
+    -------
+    carrington_map : sunpy.map.Genericmap
+        Reprojected map.
+    """
+    # The Carrington coordinate frame in sunpy assumes the standard photospheric
+    # radius for the surface of the sun, so need to temporarily set this in the
+    # map header
```
ayshih, line 44: `HeliographicCarrington` has supported the `rsun` frame attribute for a while now, and `make_fitswcs_header()` should populate `RSUN_REF` using that frame attribute.  This juggling of `RSUN_REF` should be unnecessary.  You should simply pass the `RSUN_REF` into your `carrington_header()` function to use when creating the reference coordinate.

sunkit_image/reproject.py
```diff
@@ -0,0 +1,106 @@
...
+from astropy.wcs import WCS
+from sunpy.map import Map, make_fitswcs_header
+from sunpy.sun import constants
+
+__all__ = ["reproject_carrington"]
+
+_reproject_funcs = {"interp": reproject_interp, "exact": reproject_exact, "adaptive": reproject_adaptive}
+
+
+def reproject_carrington(smap, shape_out, latitude_projection="CAR", *, reproject_method="interp", **reproject_kwargs):
```
Cadair, line 19: Given the recent addition of `GenericMap.reproject_to` should we be encouraging the use of `smap.reproject_to(carrington_heasder(...))` instead of this function?

or put a different way: what's this making easier than that?

dstansby: I think reprojecting to a Carrington frame (as opposed to a generic frame) is a common enough task that it deserves to have a really easy way to do it, ie.

```python
car_map = reproject_carrington(smap, shape_out)
```
vs.
```python
header = carrington_header(smap.date, shape_out, smap.observer_coordinate)
car_map = smap.reproject_to(WCS(header))
```

Cadair: I see your point, but why not have this API:

```python
wcs = carrington_wcs_from_map(smap, shape_out)
car_map = smap.reproject_to(wcs)
```

I feel like keeping the call to `reproject_to` maintains awareness that this is a thing you can do to any map and wcs and it simplifies the API by separating the two parts, i.e. you don't need to pass through all the reproject kwargs and the args for making the output wcs at the same time etc?

Cadair: Also as an aside: why can't we pass a header into `reproject_to`?

dstansby: I'm 👍 to that, didn't think about improving `carrington_header_from_map`. And maybe we can pass a header to `reproject_to`? I forget...

ayshih: Of course you can supply a header.  What do you take me for, a monster? =P

setup.cfg
```diff
@@ -31,9 +31,11 @@ include_package_data = True
 setup_requires =
   setuptools_scm
 install_requires =
+  reproject
   scikit-image>=0.18.0
   sunpy>=4.0.0
 
+
```
nabobalis, line 38:
```suggestion
```

examples/reproject_carrington.py
```diff
@@ -0,0 +1,40 @@
+"""
+=========================
+Creating a Carringotn Map
```
nabobalis, line 3:
```suggestion
Creating a Carrington Map
```

Cadair (commented):
I like this a lot more now. I do wonder though if it doesn't actually belong in core? We have the fits header helper in core, and this is just a more specific version of that header helper, so why is this in here?

Edit: Maybe I just haven't settled into the scope of `-image` but to me it's a home for image processing routines and this now isn't one?

sunkit_image/reproject.py
```diff
@@ -0,0 +1,65 @@
...
+    frame_out = SkyCoord(
+        0 * u.deg,
+        0 * u.deg,
+        frame="heliographic_carrington",
+        obstime=smap.date,
+        observer=smap.observer_coordinate,
+    )
+
+    if projection_code == "CAR":
+        scale = [180 / int(shape_out[0]), 360 / int(shape_out[1])] * u.deg / u.pix
```
ayshih, line 57:
```suggestion
        scale = [360 / int(shape_out[1]), 180 / int(shape_out[0])] * u.deg / u.pix
```
`scale` is specified in [horizontal, vertical], unlike `shape_out`, which is specified in [vertical, horizontal].  Everything works in your tests because you test only square pixels.

sunkit_image/reproject.py
```diff
@@ -0,0 +1,65 @@
...
+    elif projection_code == "CEA":
+        # Since, this map uses the cylindrical equal-area (CEA) projection,
+        # the spacing needs to be to 180/pi times the sin(latitude)
+        # spacing
+        # Reference: Section 5.5, Thompson 2006
+        scale = [180 / int(shape_out[0] / (np.pi * 2)), 360 / int(shape_out[1])] * u.deg / u.pix
```
ayshih, line 63:
```suggestion
        scale = [360 / int(shape_out[1]), 360 / np.pi / int(shape_out[0])] * u.deg / u.pix
```
As above, `scale` is specified in [horizontal, vertical], unlike `shape_out`, which is specified in [vertical, horizontal].

Also, the calculation for CDELT in the latitude direction was incorrect.  The size of the coordinate pixel in the sin(latitude) direction needs to be specified as 180/pi times the range of sin(latitude) (here, 2, for -1 to +1) divided by the number of sin(latitude) bins.

ayshih (changes requested):

PR #98, opened 2022-09-14
Improve doc layout

dstansby:
Given about 10 minutes of work, I think this is an improvement?


docs/dev.rst
```diff
@@ -1,7 +1,9 @@
-****************
-Formatting Style
-****************
+***********
+Development
```
nabobalis, line 2: I was planning to remove this page, the pre-commit handles all of this now.

nabobalis (approved):

nabobalis:
Not sure how I feel with the API in the sidebar uncollected.

nabobalis:
Since this is much better than what we had, I am going to merge it and probably do a follow up.

PR #123, opened 2023-02-11
Revert "Removal of ASDA and its util funtions (#118)"

nabobalis:
Adds this back with a licence update.


nabobalis:
pre-commit autofix

Issue #69, opened 2021-08-03
Inconsistent input types across functions


wtbarnes:
Most functions in this package accept the image as a bare numpy array. However, a few (e.g. `radial.fnrgf`) take a `sunpy.map.GenericMap` as an input.

It would be convenient to always accept and return a `Map`. However, this is not always practical, e.g. in the `time_lag.cross_correlation` function, it is useful to be able to provide everything from 1D timeseries to 3D image cubes to the function, none of which can be contained in a `Map` object at the moment. 

I think, in general, it would be useful to accept either a map or a bare array (e.g. using something like `singledispatch`) unless it is not practical to accept a `Map` at all (e.g. in the time lag case).

wafels:
The original idea was to have the algorithms operate on bare numpy arrays only. It insulates the image processing from changes to the `Map` object. I think this approach is still appropriate.

Cadair:
I think we should be providing map (or ndcube) aware functions for as many cases as possible, even if there are underlying functions which are implementations of things just targeting a numpy array?

wafels:
I am sure the community would like the convenience of passing in a `Map` or some higher level object. Some of the existing algorithms may need to be re-implemented to separate out the base algorithm.

Saksham-13:
Hey, I would like to work on this issue, can you point out any function which would be easy to convert than others so I can start working on it, Thanks.

nabobalis:
This issue is related to the GSoC project, as such we won't be allowing work on this for now.

PR #135, opened 2023-03-15
Accept either arrays or maps as input where appropriate

dstansby:
This PR adds a decorator that can be used to mark a function argument as accepting either array or map input. There's also a flag that can convert the output of the function to a map.

To test I've modified an existing test fixture so at least one test gets run with `ndarray` *and* `GenericMap` input for each of the modified functions.

I think I've covered all the functions where this is appropriate, but a reviewer double checking this would be good!

Fixes https://github.com/sunpy/sunkit-image/issues/69.


nabobalis (approved):

nabobalis:
Thanks again @dstansby

PR #182, opened 2024-02-05
removed marks from fixtures in tests

Ghaithq:
pytest deprecated the usage of marks with fixtures. so I removed the marks from fixtures and made sure that the marks are in the right place.

### DIFF
--- a/sunkit_image/tests/conftest.py
+++ b/sunkit_image/tests/conftest.py
@@ -155,7 +155,6 @@ def granule_minimap3():
 
 
 @pytest.fixture(params=["array", "map"])
-@pytest.mark.remote_data()
 def aia_171(request):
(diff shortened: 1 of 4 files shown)


sunkit_image/tests/conftest.py
```diff
@@ -155,7 +155,6 @@ def granule_minimap3():
 
 
 @pytest.fixture(params=["array", "map"])
-@pytest.mark.remote_data()
```
nabobalis, line 158: Unfortunately this might fail the CI as we have to make sure any tests that call these fixtures are marked with `@pytest.mark.remote_data()`

Ghaithq: I think I made sure that this is the case, but I can check again.

nabobalis (approved):

nabobalis:
Thanks for the PR @Ghaithq

Issue #186, opened 2024-02-21
Would it be better to notify the users that `calculate_solar_rotate_shift` and `mapsequence_coalign_by_rotation` mainly work on SDO cutout data?


yjzhu-solar:
### Provide a general description of the issue or problem.

I am trying to use  `calculate_solar_rotate_shift` and `mapsequence_coalign_by_rotation` to remove solar rotation in the Solar Orbiter EUI/HRI data. However, I found these functions are not designed to derotate the images except for SDO cutout data. 

For example, in `mapsequence_coalign_by_rotation`, the xshift and yshift in arcsecs are converted to pixels by 
```
xshift_keep[i] = xshift_arcseconds[i] / m.scale[0]
yshift_keep[i] = yshift_arcseconds[i] / m.scale[1]
```
which assumes a diagonal rotation matrix PC_ij. There are similar treatments in `calculate_match_template_shift` and `mapsequence_coalign_by_match_template` but they cancel each other. 

In addition, the rotation shift in arcseconds is calculated by the difference between the coordinates of the reference layer center and the derotated map center,
```
new_coordinate = solar_rotate_coordinate(m.center, observer=rotate_to_this_layer.observer_coordinate, **kwargs)
xshift_arcseconds[i] = new_coordinate.Tx - rotate_to_this_layer.center.Tx
yshift_arcseconds[i] = new_coordinate.Ty - rotate_to_this_layer.center.Ty
```

and this only works when all maps in the sequence have the same center Tx and Ty. Considering the worst scenario that the center coordinates vary significantly along the latitudinal direction, then the routine will shift the image to compensate the solar rotation and the difference in Ty.

ayshih:
Thanks!  Rather than any notification, we should fix these bugs:

> For example, in `mapsequence_coalign_by_rotation`, the xshift and yshift in arcsecs are converted to pixels by
> 
> ```
> xshift_keep[i] = xshift_arcseconds[i] / m.scale[0]
> yshift_keep[i] = yshift_arcseconds[i] / m.scale[1]
> ```
> 
> which assumes a diagonal rotation matrix PC_ij.

Yup, the code should use the PCij matrix.

> In addition, the rotation shift in arcseconds is calculated by the difference between the coordinates of the reference layer center and the derotated map center,
> 
> ```
> new_coordinate = solar_rotate_coordinate(m.center, observer=rotate_to_this_layer.observer_coordinate, **kwargs)
[...]

PR #191, opened 2024-03-28
Proposed  solution for the second Part of the issue #186 

PaoloBiolghini:
### **Proposed a  solution for the second Part of the issue #186** 

As proposed in the comments, i have used  `m.center.T..` to avoid the possible issues given by the difference between the centers  of the maps in the sequence. 
`xshift_arcseconds[i] = new_coordinate.Tx - m.center.Tx`
`yshift_arcseconds[i] = new_coordinate.Ty - m.center.Ty`

In order to make the rotation vectors relative to the reference index, before returnig these vectores i subtract the value of the recerence map to all the elements; in this way the rotation of the reference map is 0.
` xshift_arcseconds=[x-xshift_ref for x in xshift_arcseconds]`
   ` yshift_arcseconds=[y-yshift_ref for y in yshift_arcseconds]`

I have also removed the variable `rotate_to_this_layer` and created a variable `observer_coordinate_ref_layer` that stores the observer coordinate because we no longer use  `rotate_to_this_layer.center.T..`.

### DIFF
--- a/sunkit_image/coalignment.py
+++ b/sunkit_image/coalignment.py
@@ -614,20 +614,25 @@ def calculate_solar_rotate_shift(mc, layer_index=0, **kwargs):
     nt = len(mc.maps)
     xshift_arcseconds = np.zeros(nt) * u.arcsec
     yshift_arcseconds = np.zeros_like(xshift_arcseconds)
-    rotate_to_this_layer = mc.maps[layer_index]
+    observer_coordinate_ref_layer = mc.maps[layer_index].observer_coordinate
 
     for i, m in enumerate(mc):
-        # The shift of the reference layer is always zero by definition.
-        if i == layer_index:
-            continue
-
+        
         # Calculate the rotation of the center of the map 'm' at its
         # observation time to the observation time of the reference layer
         # indicated by "layer_index".
-        new_coordinate = solar_rotate_coordinate(m.center, observer=rotate_to_this_layer.observer_coordinate, **kwargs)
+        new_coordinate = solar_rotate_coordinate(m.center, observer=observer_coordinate_ref_layer, **kwargs)
 
-        xshift_arcseconds[i] = new_coordinate.Tx - rotate_to_this_layer.center.Tx
-        yshift_arcseconds[i] = new_coordinate.Ty - rotate_to_this_layer.center.Ty
+        xshift_arcseconds[i] = new_coordinate.Tx - m.center.Tx
+        yshift_arcseconds[i] = new_coordinate.Ty - m.center.Ty
+
+    # Set the shifts relative to the reference layer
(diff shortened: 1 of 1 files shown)


nabobalis:
Is there a before and after this change example or output you can show us?

PaoloBiolghini:
> Is there a before and after this change example or output you can show us?

I can try to create one if it is helpful

nabobalis:
I would like to know if this works and what it changes to the output, so I think it would be helpful.

PaoloBiolghini:
> I would like to know if this works and what it changes to the output, so I think it would be helpful.

Hi, I have create a python notebook that show the difference with the previous method
https://paolobiolghini.github.io/Sunkit_Image_Examples/showExample.html

nabobalis:
> > I would like to know if this works and what it changes to the output, so I think it would be helpful.
> 
> Hi, I have create a python notebook that show the difference with the previous method [paolobiolghini.github.io/Sunkit_Image_Examples/showExample.html](https://paolobiolghini.github.io/Sunkit_Image_Examples/showExample.html)

Thanks for the notebook, but I wasn't clear so I caused confusion.

This change did not affect any of the unit tests and that means we do not have a way to validate this change within our test suite. We can't merge a change like this without some sort of test to ensure that it was broken before, its fixed now and will be fixed in the future.

Could you add a unit test for this?

PaoloBiolghini:
> > > I would like to know if this works and what it changes to the output, so I think it would be helpful.
> > 
> > 
> > Hi, I have create a python notebook that show the difference with the previous method [paolobiolghini.github.io/Sunkit_Image_Examples/showExample.html](https://paolobiolghini.github.io/Sunkit_Image_Examples/showExample.html)
> 
> Thanks for the notebook, but I wasn't clear so I caused confusion.
> 
> This change did not affect any of the unit tests and that means we do not have a way to validate this change within our test suite. We can't merge a change like this without some sort of test to ensure that it was broken before, its fixed now and will be fixed in the future.
> 
> Could you add a unit test for this?

yes i can try in the next few days

PaoloBiolghini:
is it possible to use my custom map inside the tests or should i use the one that are already used in the tests?

nabobalis:
I would prefer if we can use existing ones however, if the existing ones are not good enough to use. Then we can consider adding some resized test data that works

PaoloBiolghini:
is actually requiring more time than i thought because i have some troubles with pytest

nabobalis:
That's ok, there is no deadline or rush. 

Anything I can try to help with?

wtbarnes:
Thanks for your efforts here @PaoloBiolghini. Given the complete refactoring of the API that's ongoing in #207, I'm going to close this PR.

Issue #96, opened 2022-09-06
Failing test `test_multiscale_gaussian`


nabobalis:
### Describe the bug

`test_multiscale_gaussian` is failing sometimes on the CI.

I have been unable to replicate any of the fails locally using tox, changing Python version etc.

I have marked it as `xfail` for now but we need to debug this, see https://github.com/sunpy/sunkit-image/pull/95/

### To Reproduce

I can't.

meghana151:
Hello, Can I work on this?

nabobalis:
Sure.

Deus1704:
I have tested by removing the @pytest.mark.xfail for the `test_multiscale_gaussian` with the python versions: 3.9, 3.10, 3.11 and their corresponding oldest dependencies too and all of them pass.

I believe that this issue can be safely closed.

nabobalis:
Can you pull request the removal of the mark? Only if it passes on the ci, will we remove this

Deus1704:
sure!

PR #193, opened 2024-03-30
removed the xfail mark

Deus1704:
## PR Description
This pull request aims to assess whether removing "xfail" results in the failure of the mgn unit test within the CI.
Related to #96 
Also since this is an internal test, changelog won't be required

### DIFF
--- a/sunkit_image/tests/test_enhance.py
+++ b/sunkit_image/tests/test_enhance.py
@@ -21,7 +21,6 @@ def map_test():
     return np.ones((4, 4), dtype=float)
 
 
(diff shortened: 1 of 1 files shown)


Deus1704:
I think the dependencies aren't being correctly installed in the py311-devdeps, which is causing the entire test suite to fail

nabobalis:
Yes, I will fix that. 

Thanks for the PR @Deus1704

nabobalis (approved):

Deus1704:
Glad I could be of help!

Issue #203, opened 2024-05-31
Remove `calculate_solar_rotate_shift` and `mapsequence_coalign_by_rotation`


wtbarnes:
### Provide a general description of the issue or problem.

These methods are not needed and are just worse versions of functionality we already have in `sunpy` to reproject images given some model of differential rotation, e.g. https://docs.sunpy.org/en/stable/generated/gallery/differential_rotation/reprojected_map.html#sphx-glr-generated-gallery-differential-rotation-reprojected-map-py

Deus1704:
This issue too can be marked complete since the #206 has been merged.

PR #206, opened 2024-06-04
Removes coalign by solar rotation completely

Deus1704:
## PR Description
#203 
Following our discussion on the removal of the methods `calculate_solar_rotate_shift` and `mapsequence_coalign_by_rotation`, this PR removes them along with removal of the example gallery and the unit tests.


nabobalis (approved):

wtbarnes:
The doc build fail is real. It looks like there are still a few references to the removed functions somewhere in the docs. I'm not sure what's going on with the online build.

changelog/206.removal.rst
```diff
@@ -0,0 +1,5 @@
+The coalignment methods ``calculate_solar_rotate_shift`` and ``mapsequence_coalign_by_rotation`` present in the `sunkit_image.coalignment` have been removed, with no replacement.
+Correspondingly the gallery example ``Compensating for Solar Rotation`` making use of these methods are removed.
+This is because they are worse versions of the functionality already present in sunpy to reproject images given some model of differential rotation.
```
wtbarnes, line 3: This is not necessary.

```suggestion
```

wtbarnes (changes requested):
As noted above, the doc build failure still needs fixing.

Deus1704:
> As noted above, the doc build failure still needs fixing.

Yeah there were few references in the legacy changelog dating back to the initial addition of these functions.

nabobalis:
Online test is a file download failure.

Deus1704:
> Online test is a file download failure.

Yeah, but I believe we should look into it? because the same issue is occurring when I run the online test suite locally and the file being downloaded is a sunpy hosted sample data, so ideally this failure shouldn't occur?

nabobalis:
> > Online test is a file download failure.
> 
> Yeah, but I believe we should look into it? because the same issue is occurring when I run the online test suite locally and the file being downloaded is a sunpy hosted sample data, so ideally this failure shouldn't occur?

Yeah but it isnt due to this PR so we can gloss over it for this.

Issue #211, opened 2024-06-27
Test code with cutouts


nabobalis:
### Describe the bug

Cutouts seem to change the DTYPE and we need to make sure our code works with them. 

The bug report was on Element here:
https://matrix.to/#/!MeRdFpEonLoCwhoHeT:matrix.org/$171951269810519jrmom:matrix.org?via=matrix.org&via=openastronomy.org&via=cadair.com

PR #215, opened 2024-06-28
Add tests with a jsoc cutout file and update MGN to convert int to float

nabobalis:
Fixes https://github.com/sunpy/sunkit-image/issues/211

Issue was reported with JSOC cutouts and the datatype breaking these routines.


sunkit_image/enhance.py
```diff
@@ -101,51 +101,42 @@ def mgn(
...
     image = np.zeros_like(data)
     conv = np.zeros_like(data)
     sigmaw = np.zeros_like(data)
-
     for s, weight in zip(sigma, weights, strict=True):
         # 2 & 3 Create kernel and convolve with image
         # Refer to equation (1) in the paper
-        ndimage.gaussian_filter(data, sigma=s, truncate=truncate, mode="nearest", output=conv)
-
+        conv = ndimage.gaussian_filter(data, sigma=s, truncate=truncate, mode="nearest")
```
wtbarnes, line 116: Why the change here? Presumably the `output` kwarg is used to avoid making a copy of the array.

sunkit_image/enhance.py
```diff
@@ -101,51 +101,42 @@ def mgn(
...
         )
-
     if weights is None:
         weights = np.ones(len(sigma))
-
     # 1. Replace spurious negative pixels with zero
     if clip is True:
         data[data <= 0] = 1e-15  # Makes sure that all values are above zero
-
+    data = data.astype(np.float32) if data.dtype != np.float32 else data
```
wtbarnes, line 109: Why float32? Why not double precision for example?

sunkit_image/enhance.py
```diff
@@ -101,51 +101,42 @@ def mgn(
...
         # square the difference, and convolve with kernel. Square-root the
         # resulting image to give `local standard deviation` image sigmaw
         # Refer to equation (2) in the paper
         conv = data - conv
-        ndimage.gaussian_filter(conv**2, sigma=s, truncate=truncate, mode="nearest", output=sigmaw)
-        np.sqrt(sigmaw, out=sigmaw)
-
+        sigmaw = ndimage.gaussian_filter(conv**2, sigma=s, truncate=truncate, mode="nearest")
+        with np.errstate(invalid="ignore"):
+            sigmaw = np.sqrt(sigmaw)
```
wtbarnes, line 124: Same comment as above. Why not make use of `output`?

sunkit_image/enhance.py
nabobalis: I just picked one at random. I should just promote it to that instead. That makes more sense.

sunkit_image/enhance.py
nabobalis: I just don't like it but yeah. I will undo this.

sunkit_image/enhance.py
nabobalis: Trying that I still get:

```
numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'sqrt' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

I can probably make it unsafe?

nabobalis: I came up with worse ideas.

sunkit_image/asda.py
```diff
@@ -1,391 +1,391 @@
-"""
```
nabobalis, line 1: This had the wrong last line character.

sunkit_image/granule.py
```diff
@@ -1,282 +1,282 @@
-"""
```
nabobalis, line 1: This had the wrong last line character.

sunkit_image/tests/test_granule.py
```diff
@@ -1,124 +1,124 @@
-import numpy as np
```
nabobalis, line 1: This had the wrong last line character.

alasdairwilson (approved):

alasdairwilson:
I don't count but +1

wtbarnes:
What was the actual issue here? Do the cutouts from the JSOC have a different datatype than the full-disk L1 files?

nabobalis:
> What was the actual issue here? Do the cutouts from the JSOC have a different datatype than the full-disk L1 files?

I didn't debug the reason.

sunkit_image/tests/test_granule.py
nabobalis, line 1: This is jut a end of file character change.

sunkit_image/asda.py
nabobalis, line 1: This is jut a end of file character change.

sunkit_image/enhance.py
```diff
@@ -101,51 +102,40 @@ def mgn(
...
         # Refer to equation (1) in the paper
         ndimage.gaussian_filter(data, sigma=s, truncate=truncate, mode="nearest", output=conv)
-
         # 4. Calculate difference between image and the local mean image,
         # square the difference, and convolve with kernel. Square-root the
         # resulting image to give `local standard deviation` image sigmaw
         # Refer to equation (2) in the paper
         conv = data - conv
         ndimage.gaussian_filter(conv**2, sigma=s, truncate=truncate, mode="nearest", output=sigmaw)
         np.sqrt(sigmaw, out=sigmaw)
```
nabobalis, line 123:
```python
>           np.sqrt(sigmaw, out=sigmaw)
E           numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'sqrt' output from dtype('float32') to dtype('int16') with casting rule 'same_kind'
```
So this is the original issue that we need to fix.

wtbarnes (approved):

PR #225, opened 2024-08-01
Files had wrong char ending

nabobalis:
Fixed now to use linux ones.


wtbarnes:
Sorry, what in the world is happening here?

nabobalis:
Text files created on DOS/Windows machines have different line endings than files created on Unix/Linux. DOS uses carriage return and line feed ("\r\n") as a line ending, which Unix uses just line feed ("\n").

PR #266, opened 2025-02-20
Update radial.py to include "none" as an option for rhef

GillySpace27:
## PR Description

If one wants to use the upsilon correction without the RHEF, they can select "none" as an option. This also implements casefold on the method decision tree and sets the default sorting algorithm to be the scipy rankdata.


GillySpace27:
This PR addresses Issue #267

nabobalis (approved):

PR #321, opened 2026-06-11
Fix import for deprecation on scikit-image>0.20

samaloney:
## PR Description

Try old import and fail back to new import can be removed once scikit-image>0.20

## AI Assistance Disclosure

AI tools were used for:
- [  ] Code generation (e.g., when writing an implementation or fixing a bug)
- [  ] Test/benchmark generation
- [  ] Documentation (including examples)
- [  ] Research and understanding
- [x ] No AI tools were used

> Regardless of AI use, the human contributor remains fully responsible for correctness, design choices, licensing compatibility, and long-term maintainability.

### DIFF
--- a/changelog/321.trivial.rst
+++ b/changelog/321.trivial.rst
@@ -1,0 +1,1 @@
+Update imports in ``stara`` to account for change in scikit-image>0.20.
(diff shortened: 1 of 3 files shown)


alasdairwilson (approved):

alasdairwilson:
hmm, I approved because it fixes the deprecation but the default behaviour has changed, the mode before used to default to "reflect" but now it is "ignore" so since we never se tmode our behaviour is changing from reflect to ignore so perhaps mode="reflect" should be included

unless we dont care.

PR #325, opened 2026-06-12
Fast path for find_pixel_radii on non-rotated HPC maps

GillySpace27:
## Motivation

After #324 collapsed the inner loop in `~sunkit_image.radial.rhef`, profiling showed `~sunkit_image.utils.find_pixel_radii` had become the new bottleneck: on a 4096² map it costs ~1 s per call, and `rhef` calls it twice in succession (once via `find_radial_bin_edges`, once via `blackout_pixels_above_radius`). That is ~2 s of WCS conversion before any filtering happens, on what is functionally a fixed geometric transform.

The slow path uses `smap.wcs.pixel_to_world` and the full SkyCoord stack for every pixel, but for the common case — a non-rotated helioprojective-Cartesian map with a known `rsun_obs` — the answer is a closed-form function of CDELT, CRPIX, CRVAL, and the gnomonic projection. This PR adds that fast path (and lets internal callers avoid the duplicate call).

## Summary

Two coupled changes:

1. **Analytic fast path for `find_pixel_radii`**: when both axes are HPLN/HPLT and the rotation matrix is the identity, compute pixel-to-Sun-centre distance directly from `cdelt`, `crpix`, `crval`, and `rsun_obs` plus the gnomonic-TAN inverse. Rotated or non-HPC maps fall through to the existing SkyCoord path unchanged.

2. **Optional `map_r=` kwarg on `blackout_pixels_above_radius`**: callers that already hold the radii array can pass it in and skip the internal `find_pixel_radii` call. `rhef` now does this: it has `map_r` from `find_radial_bin_edges` and threads it through, eliminating the duplicate WCS pass.

## Equivalence

The fast-path output matches the SkyCoord path to:
- **better than 2e-4 R⊙** on a 4096² map at 2"/px (the worst-case scenario; the diff is concentrated in the corner pixels)
- **floating-point precision** on AIA-scale 0.6"/px maps

Tests cover:
- Equivalence on sun-centred, off-centre (`CRVAL != 0`), and `scale=`-rescaled cases.
- Non-identity rotation triggers the SkyCoord fallback (verified by mocking the fast helper).
- `blackout_pixels_above_radius(map_r=...)` produces byte-equal output to the no-kwarg call, and the patched internal `find_pixel_radii` is verified NOT to be called.

## Performance

Measured on a synthetic 4096² HPLN-TAN map, macOS arm64, numpy 2.0, astropy 7.x:

|  | call time | speedup |
|---|---|---|
| Slow path (`smap.wcs.pixel_to_world`) | 1.05 s | 1× |
| Fast path (this PR) | 0.15 s | **~7×** |

For `rhef` specifically — which calls `find_pixel_radii` twice — the WCS-bound portion drops from ~2.1 s to ~150 ms total. Combined with the inner-loop rewrite in #324, a 4096² × 720-bin `rhef` should run in around 250 ms instead of the original ~11 s.

## Notes

- No public API changes other than the new keyword-only `map_r=` argument on `blackout_pixels_above_radius`.
- The towncrier fragment `changelog/325.feature.rst` covers both speedup mechanisms in a single paragraph.
- This PR is branched from `main`, parallel to #324; they touch `sunkit_image/radial.py` in different spots and one will need a trivial rebase after the first lands. I'm happy to do that — just let me know merge order.
- Like #324, this also carries a small drive-by fix pinning `skimage.morphology.white_tophat`'s `mode='reflect'` argument in `sunkit_image/stara.py`. The CI matrix `py314-devdeps` job was failing on `test_stara` due to a `PendingSkimage2Change` deprecation that's unrelated to the rest of this PR; fix is conditional on kwarg availability so it stays compatible with the project's minimum scikit-image (0.20). Happy to split into its own PR.

cc @sunpy/sunkit-image-maintainers


sunkit_image/utils/utils.py
```diff
@@ -94,6 +94,51 @@ def bin_edge_summary(r, binfit):
...
+    Only valid when ``_is_simple_hpc(smap)`` is True; callers must check.
+    """
+    ny, nx = smap.data.shape
+    cdelt = smap.wcs.wcs.cdelt
+    crpix = smap.wcs.wcs.crpix - 1.0  # FITS->numpy index origin
+    crval = smap.wcs.wcs.crval
+    cunit = smap.wcs.wcs.cunit
+    # Sky-plane offsets in arcsec (broadcast across the 2-D grid).
+    dx_as = ((np.arange(nx) - crpix[0]) * cdelt[0] + crval[0]) * u.Unit(cunit[0]).to(u.arcsec)
+    dy_as = ((np.arange(ny) - crpix[1]) * cdelt[1] + crval[1]) * u.Unit(cunit[1]).to(u.arcsec)
```
ayshih, line 131: Uses the small-angle approximation for the WCS projection, and assuming that the reference coordinate is close to Sun center

sunkit_image/utils/utils.py
```diff
@@ -94,6 +94,51 @@ def bin_edge_summary(r, binfit):
...
+    r_planar = np.sqrt(dx_as[None, :] ** 2 + dy_as[:, None] ** 2)
```
ayshih, line 132: Uses the small-angle approximation

sunkit_image/utils/utils.py
```diff
@@ -94,6 +94,51 @@ def bin_edge_summary(r, binfit):
...
+    return np.allclose(smap.rotation_matrix, np.eye(2))
+
+
+def _find_pixel_radii_fast(smap, scale):
+    """
+    Analytic fast path for `find_pixel_radii`.
+
+    Skips ``smap.wcs.pixel_to_world``'s SkyCoord round-trip and computes the
+    pixel-to-sun-centre distance directly from the WCS header values.  Uses
+    the TAN gnomonic projection's spherical-distance inverse so the output
```
ayshih, line 119: Given the approximations being used, this calculation is not actually specific to the gnomonic projection, but the inaccuracy will depend on the projection

ayshih (commented):
I'm uneasy with the level of approximation in this PR because the inaccuracies will become unacceptable off the solar disk.  (The inaccuracies would be "okay" if `find_pixel_radii()` were only about identifying the solar limb, but the function is more general than that.)  My inclination would be for `find_pixel_radii()` to not use this fast path automatically, but rather have the user opt-in via a keyword argument.

GillySpace27:
Good point — thanks. The fast path does lean on the small-angle / reference-near-Sun-center assumptions, and `find_pixel_radii()` is general enough (used well off-disk too) that silently swapping in the approximation is the wrong default.

I've pushed a commit making it opt-in via a keyword-only `fast=` argument (default `False`):

- `fast=False` (the default) always runs the exact SkyCoord-based path — no behavior change for any existing caller.
- `fast=True` uses the analytic path only when the map is non-rotated HPC; rotated and non-HPC maps still fall through to the exact path even when `fast=True`.
- The docstring now spells out that the fast path is an approximation whose accuracy degrades away from disk center, and there's a new test asserting the default never touches `_find_pixel_radii_fast`.

Let me know if you'd prefer a different keyword name or a warning when `fast=True` silently falls through.

